PR #3, opened 2019-09-09
Attempt to ensure that watch is always sequential

ibuildthecloud:
This is currently untested, but the idea is to ensure that watch events are always sequential to avoid missing an ID due to an out of order insert.  In such a situation that we see a gap in the log we give up to 1 second to find the missing record.  If it is not discovered in 1 second we move on with the log.

### DIFF
--- a/pkg/logstructured/sqllog/sql.go
+++ b/pkg/logstructured/sqllog/sql.go
@@ -37,6 +37,8 @@ type Dialect interface {
 	DeleteRevision(ctx context.Context, revision int64) error
 	GetCompactRevision(ctx context.Context) (int64, error)
 	SetCompactRevision(ctx context.Context, revision int64) error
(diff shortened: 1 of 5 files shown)


pkg/logstructured/sqllog/sql.go
```diff
@@ -266,10 +266,12 @@ func (s *SQLLog) startWatch() (chan interface{}, error) {
...
 func (s *SQLLog) poll(result chan interface{}) {
 	var (
-		last int64
+		last     int64
+		skip     int64
+		skipTime time.Time
 	)
 
-	wait := time.NewTicker(120 * time.Second)
+	wait := time.NewTicker(time.Second)
```
yue9944882: this reminds of the previous discussion about making the value configurable.. but 1 second sounds like a good magic number to me for now

PR #19, opened 2019-12-20
Update function signature on nocgo stub

carlosedp:
Fix signature of `sqlite.New` function updated on https://github.com/rancher/kine/commit/c395d22fffc82863dccd39bb807a26bafd107b9d.


carlosedp:
@ibuildthecloud any tips on how can this be tested for both cases (with and without cgo) to avoid this from happening in the future?

PR #21, opened 2020-01-17
Use correct error types so client can properly handle errors

ibuildthecloud:


brandond:
Surprised to see `ErrGRPCFutureRev` ever returned.

PR #25, opened 2020-01-29
Fix standard linter errors

daxmc99:
Cleanup some small linter issues found by golangci-lint

### DIFF
--- a/pkg/logstructured/logstructured.go
+++ b/pkg/logstructured/logstructured.go
@@ -33,7 +33,9 @@ func (l *LogStructured) Start(ctx context.Context) error {
 	if err := l.log.Start(ctx); err != nil {
 		return err
 	}
-	l.Create(ctx, "/registry/health", []byte(`{"health":"true"}`), 0)
+	if _,err := l.Create(ctx, "/registry/health", []byte(`{"health":"true"}`), 0); err != nil {
+		return err
+	}
 	go l.ttl(ctx)
 	return nil
 }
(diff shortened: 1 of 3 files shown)


pkg/logstructured/logstructured.go
```diff
@@ -33,7 +33,9 @@ func (l *LogStructured) Start(ctx context.Context) error {
 	if err := l.log.Start(ctx); err != nil {
 		return err
 	}
-	l.Create(ctx, "/registry/health", []byte(`{"health":"true"}`), 0)
```
ibuildthecloud: We specifically don't care about the error here.  You can add a comment or _, _ := to indicate as such

pkg/logstructured/logstructured.go
```diff
@@ -302,7 +304,8 @@ func (l *LogStructured) ttl(ctx context.Context) {
 				return
 			case <-time.After(time.Duration(event.KV.Lease) * time.Second):
 			}
-			l.Delete(ctx, event.KV.Key, event.KV.ModRevision)
+			_,_,_,err := l.Delete(ctx, event.KV.Key, event.KV.ModRevision)
```
ibuildthecloud: In an HA setup this might cause a bunch of spurious errors.  You'd have to double check that a delete of a something that doesn't exist will thrown an error.

daxmc99: What would be a good way to check that?

cjellick:
so we dont run go lint on this project? do we do that by defualt on other projects?

daxmc99:
@cjellick This project currently has no CI enabled... I feel like I just volunteered to add CI 😅

galal-hussein:
fixed by https://github.com/rancher/kine/pull/42

PR #28, opened 2020-02-11
Minimal sample and minor tweak to listener

ibrokethecloud:
The PR contains the following:

1. A sample minimal config to use kine with rke based on a external etcd endpoint served by kine.

2. Kine endpoint is not secured using tls. There is a change to endpoints.go to use the same cert used for communication with a database to secure kine endpoint. 

If needed I can tweak the code to add extra flags to specify a different cert for kine endpoint. This would be more beneficial as this will allow users to use two sets of certs for encrypting backend and endpoint traffic.


erikwilson:
thanks for the PR @ibrokethecloud! I think the code change is okay, altho we may want to gate the behavior behind the flag.

my personal feeling is that examples should not contain static credentials, so it would be better to include a script to generate them.

the commits should be "fixed up"/rebased to two commits, one for the code change and one for the sample.

feel free to ping myself or @cjellick, in PRs or slack, if it helps to get this stuff moving.

PR #33, opened 2020-05-14
fixed invalid endpoint error

MonzElmasry:
the pr basically fixed two things:

- error out when passing dsn without " (driver) :// "
-  timeout with postgres case when it takes forever to do a db ping

And add drone automation 

https://github.com/rancher/k3s/issues/1668

### DIFF
--- a/.golangci.json
+++ b/.golangci.json
@@ -1,0 +1,19 @@
+{
+	"linters": {
+		"disable-all": true,
+		"enable": [
+			"govet",
+			"golint",
+			"goimports",
+			"misspell",
+			"ineffassign",
+			"gofmt"
+		]
+	},
+	"run": {
+		"skip-files": [
+			"/zz_generated_"
+		],
+		"deadline": "5m"
+	}
+}
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,0 +1,29 @@
+FROM golang:1.13.4-alpine3.10
+
+ARG DAPPER_HOST_ARCH
+ENV ARCH $DAPPER_HOST_ARCH
+
+RUN apk -U add bash git gcc musl-dev docker vim less file curl wget ca-certificates
+RUN go get -d golang.org/x/lint/golint && \
+    git -C /go/src/golang.org/x/lint/golint checkout -b current 06c8688daad7faa9da5a0c2f163a3d14aac986ca && \
+    go install golang.org/x/lint/golint && \
+    rm -rf /go/src /go/pkg
+RUN mkdir -p /go/src/golang.org/x && \
+    cd /go/src/golang.org/x && git clone https://github.com/golang/tools && \
+    git -C /go/src/golang.org/x/tools checkout -b current aa82965741a9fecd12b026fbb3d3c6ed3231b8f8 && \
+    go install golang.org/x/tools/cmd/goimports
+RUN rm -rf /go/src /go/pkg
+RUN if [ "${ARCH}" == "amd64" ]; then \
+        curl -sL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s v1.15.0; \
+    fi
+
+
+ENV DAPPER_ENV REPO TAG DRONE_TAG
+ENV DAPPER_SOURCE /go/src/github.com/rancher/kine/
+ENV DAPPER_OUTPUT ./bin ./dist
+ENV DAPPER_DOCKER_SOCKET true
+ENV HOME ${DAPPER_SOURCE}
+WORKDIR ${DAPPER_SOURCE}
+
+ENTRYPOINT ["./scripts/entry"]
+CMD ["ci"]
--- a/pkg/endpoint/endpoint.go
+++ b/pkg/endpoint/endpoint.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/rancher/kine/pkg/drivers/dqlite"
@@ -34,7 +35,6 @@ type Config struct {
 
 	tls.Config
 }
-
 type ETCDConfig struct {
 	Endpoints   []string
 	TLSConfig   tls.Config
@@ -42,7 +42,7 @@ type ETCDConfig struct {
 }
 
 func Listen(ctx context.Context, config Config) (ETCDConfig, error) {
-	driver, dsn := ParseStorageEndpoint(config.Endpoint)
+	driver, dsn, err := ParseStorageEndpoint(config.Endpoint)
 	if driver == ETCDBackend {
 		return ETCDConfig{
 			Endpoints:   strings.Split(config.Endpoint, ","),
@@ -50,12 +50,45 @@ func Listen(ctx context.Context, config Config) (ETCDConfig, error) {
 			LeaderElect: true,
 		}, nil
 	}
+	if err != nil {
+		return ETCDConfig{}, errors.Wrap(err, "parsing storage point")
+	}
 
-	leaderelect, backend, err := getKineStorageBackend(ctx, driver, dsn, config)
+	type Response struct {
+		leaderElect bool
+		backend     server.Backend
+		err         error
+	}
+	ch := make(chan Response, 1)
+
+	go func() {
+		leaderElect, backend, err := getKineStorageBackend(ctx, driver, dsn, config)
+		res := Response{
+			leaderElect: leaderElect,
+			backend:     backend,
+			err:         err,
+		}
+		ch <- res
+	}()
+
+	var (
+		leaderElect bool
+		backend     server.Backend
+	)
+	const storageDriveTimeout = time.Second * 10
+	select {
+	case res := <-ch:
+		leaderElect = res.leaderElect
+		backend = res.backend
+		err = res.err
+
+	case <-time.After(storageDriveTimeout):
+		return ETCDConfig{}, errors.New("getKineDriver timed out, please check your connection/host port number")
+	}
+
 	if err != nil {
 		return ETCDConfig{}, errors.Wrap(err, "building kine")
 	}
-
 	if err := backend.Start(ctx); err != nil {
 		return ETCDConfig{}, errors.Wrap(err, "starting kine backend")
 	}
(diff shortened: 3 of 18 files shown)


galal-hussein:
@MonzElmasry you might want to run `gofmt` against the whole code to fix the indentation and spaces for the code

cjellick:
@MonzElmasry what happens if the url is technically correct but points to nothing, like the database doesnt actually exist or the network is down.

In this scenario, I was expecting k3s to actually fail to start, though I'm not even sure if that is the right behavior.  @ibuildthecloud should k3s fail to start or sit in an endless retry loop waiting for the db connection to be successful? Should we introduce some sort of connection test?

@MonzElmasry while we wait for @ibuildthecloud's response, can you investigate around making a test connection? Just some research and a prototype. I think there may be some established best practices around testing connections to sql dbs, but dont know for sure.

cjellick:
@MonzElmasry I'd also like you to consider how we could potentially write tests for your changes. To my knowledge, kine doesnt currently have meaningful tests and I haven't put much thought into how we could introduce them. Maybe we pull the framework we have in k3s for hooking k3s up to containerized databases into this repository and start with some simple smoke tests of connecting to the db backends.

davidnuzik:
So when we first talked about this GH issue I believe that yes we were suggesting that K3s should NOT start at all on the initial connection to the DB. If that connection fails, K3s should not run and a sensible error should be shown to the user.

Dockerfile.dapper
```diff
@@ -0,0 +1,32 @@
...
+RUN mkdir -p /go/src/golang.org/x && \
+    cd /go/src/golang.org/x && git clone https://github.com/golang/tools && \
+    git -C /go/src/golang.org/x/tools checkout -b current aa82965741a9fecd12b026fbb3d3c6ed3231b8f8 && \
+    go install golang.org/x/tools/cmd/goimports
+RUN rm -rf /go/src /go/pkg
+RUN if [ "${ARCH}" == "amd64" ]; then \
+        curl -sL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s v1.15.0; \
+    fi
+
+RUN curl -L https://github.com/rancher/rke/releases/download/v1.0.5/rke_linux-$ARCH \
```
galal-hussein, line 20: this tool is not needed in kine

Dockerfile.dapper
```diff
@@ -0,0 +1,32 @@
...
+    --output /usr/local/bin/rke && chmod +x /usr/local/bin/rke
+
+ENV GO111MODULE off
+ENV DAPPER_ENV REPO TAG DRONE_TAG
+ENV DAPPER_SOURCE /go/src/github.com/rancher/helm-controller/
```
galal-hussein, line 25: this also need to be changed to rancher/kine

pkg/endpoint/endpoint.go
```diff
@@ -34,28 +35,58 @@ type Config struct {
...
+
+	type Response struct {
+		leaderelect bool
+		backend     server.Backend
+		err         error
+	}
+	ch := make(chan Response, 1)
+
+	go func() {
+		leaderelect, backend, err := getKineStorageBackend(ctx, driver, dsn, config)
```
briandowns, line 66:
```suggestion
		leaderElect, backend, err := getKineStorageBackend(ctx, driver, dsn, config)
```

pkg/endpoint/endpoint.go
```diff
@@ -34,28 +35,58 @@ type Config struct {
...
+		res := Response{leaderelect: leaderelect, backend: backend, err: err}
+		ch <- res
+	}()
+
+	var (
+		leaderelect bool
```
briandowns, line 72:
```suggestion
		leaderElect bool
```

pkg/endpoint/endpoint.go
```diff
@@ -34,28 +35,58 @@ type Config struct {
...
+	)
+
+	select {
+	case res := <-ch:
+		leaderelect = res.leaderelect
+		backend = res.backend
+		err = res.err
+
+	case <-time.After(10 * time.Second):
+		return ETCDConfig{}, fmt.Errorf("Timeout, please check your connection/ host port number")
```
briandowns, line 83:
```suggestion
		return ETCDConfig{},errors.New("Timeout, please check your connection/ host port number")
```

pkg/endpoint/endpoint.go
```diff
@@ -34,28 +35,58 @@ type Config struct {
...
+	type Response struct {
+		leaderelect bool
+		backend     server.Backend
+		err         error
+	}
+	ch := make(chan Response, 1)
+
+	go func() {
+		leaderelect, backend, err := getKineStorageBackend(ctx, driver, dsn, config)
+		res := Response{leaderelect: leaderelect, backend: backend, err: err}
```
briandowns, line 67: This is a little more readable if split to multiple lines.

pkg/endpoint/endpoint_test.go
```diff
@@ -0,0 +1,36 @@
...
+		{"postgres://123", "Timeout, please check your connection/ host port number"},
+		{"postgres://", "connection refused"},
+		{"text", "invalid endpoint value"},
+	}
+	for i := range endpoints {
+		config.Endpoint = endpoints[i][0]
+		_, err := Listen(ctx, config)
+		if err == nil {
+			t.Error("unpassed error !! ", endpoints[i][1])
+		} else {
```
briandowns, line 32: Is this else necessary?

pkg/endpoint/endpoint.go
```diff
@@ -138,17 +167,21 @@ func getKineStorageBackend(ctx context.Context, driver, dsn string, cfg Config)
 	return leaderElect, backend, err
 }
 
-func ParseStorageEndpoint(storageEndpoint string) (string, string) {
+func ParseStorageEndpoint(storageEndpoint string) (string, string, error) {
 	network, address := networkAndAddress(storageEndpoint)
+	var err error = nil
```
briandowns, line 172:
```suggestion
	var err error
```

pkg/endpoint/endpoint_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"context"
+	"fmt"
+	"testing"
+)
+
+func TestListen(t *testing.T) {
+	var config Config
+	var ctx context.Context
+	config.Listener = ""
+	endpoints := [13][2]string{{"postgres://doadmin:ax4t3h9twe04fa7k@postgressdb-do-user-898766-0.a.db.ondigitalocean.com:2560/defaultdb", "invalid port number"},
```
galal-hussein, line 13: This is not the right approcah for unit testing for this function, basically we want to make sure that the function returns an error when an invalid address is passed, so there are basically 3 cases:

- DSN with no <driver>:// and this should return the new error you implemented
- DSN with <driver>:// and invalid address and this should return an error
- DSN with correct <driver>:// and address

galal-hussein: you dont have to have a working mysql and postgres to test the last case, we basically have these integration tests in k3s https://github.com/rancher/k3s/blob/master/scripts/test-setup-sonobuoy-mysql

cc @cjellick

.golangci.json
```diff
@@ -0,0 +1,19 @@
+{
```
cjellick, line 1: I don think this file should be checked in, should it?

MonzElmasry: it should be added 
https://github.com/rancher/k3s/blob/master/.golangci.json
because it's used in golint-ci in drone.

davidnuzik:
@ibuildthecloud can we get final review and LGTM from you on this PR? cc: @cjellick

ibuildthecloud:
Can this PR be split into two commits.  All the CI changes in one and the code changes in another.  Also, don't delete Dockerfile, that file is used.

pkg/endpoint/endpoint.go
```diff
@@ -34,28 +35,60 @@ type Config struct {
...
 			Endpoints:   strings.Split(config.Endpoint, ","),
 			TLSConfig:   config.Config,
 			LeaderElect: true,
 		}, nil
 	}
+	if err != nil {
+		return ETCDConfig{}, errors.Wrap(err, "parsing storage point")
+	}
+
+	type Response struct {
```
ibuildthecloud, line 57: Sorry, I don't fully understand the context, but it really doesn't seem like the right place to actually be testing the connection.  What is the issue you are trying to fix that requires testing the connection?

pkg/endpoint/endpoint.go
```diff
@@ -138,17 +169,21 @@ func getKineStorageBackend(ctx context.Context, driver, dsn string, cfg Config)
 	return leaderElect, backend, err
 }
 
-func ParseStorageEndpoint(storageEndpoint string) (string, string) {
+func ParseStorageEndpoint(storageEndpoint string) (string, string, error) {
 	network, address := networkAndAddress(storageEndpoint)
+	var err error
```
ibuildthecloud, line 174: Is it more straightfoward to just say
```go
if storageEndpoint == "" {
  return SQLiteBackend, ""
}
```
And change nothing else?

pkg/endpoint/endpoint_test.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"context"
+	"testing"
+)
+
+func TestListen(t *testing.T) {
+	var config Config
+	var ctx context.Context
+	config.Listener = ""
+	endpoints := [5]string{
```
ibuildthecloud, line 12:
```suggestion
	endpoints := []string{
```
The 5 is not needed.  Putting a number there creates an actual array and not a slice.  Arrays are using in go really only in very performance critical situations.

pkg/endpoint/endpoint_test.go
```diff
@@ -0,0 +1,26 @@
+package endpoint
+
+import (
+	"context"
+	"testing"
+)
+
+func TestListen(t *testing.T) {
+	var config Config
+	var ctx context.Context
```
ibuildthecloud, line 10:
```suggestion
	var (
		config Config
		ctx context.Context
	)
```
Just looks more idiomatic this way

pkg/endpoint/endpoint_test.go
```diff
@@ -0,0 +1,26 @@
...
+	config.Listener = ""
```
ibuildthecloud, line 11: Is this line redundant?  Strings are initialized to "" by default.

PR #34, opened 2020-05-18
Flag to enable TLS on kine endpoint

ibrokethecloud:
By default kine endpoint does not have tls encryption.

The PR contains minor tweaks to the endpoint Config struct, to allow tls behaviour to be gated.

The default behaviour is to have no tls encryption on the endpoint.

This should allow existing kine uses cases to work without tls, while allowing kine to be used with additional k8s provisioning tools like kubeadm.


galal-hussein:
@ibrokethecloud Can you squash the commits, and will LGTM. thanks!

ibrokethecloud:
@galal-hussein sorted it out.

pkg/endpoint/endpoint.go
```diff
@@ -62,14 +64,14 @@ func Listen(ctx context.Context, config Config) (ETCDConfig, error) {
 
 	listen := config.Listener
 	if listen == "" {
-		listen = KineSocket
+		listen = "KineSocket"
```
brandond, line 67: This is not right - `KineSocket` is a variable, why did you quote it?
```go
KineSocket      = "unix://kine.sock"
```

ibrokethecloud: The best reason i can think of is a typo 🤦

brandond:
Server-side TLS support was added in https://github.com/k3s-io/kine/pull/103

PR #39, opened 2020-06-30
Fix --endpoint argument for sqlite backend (#38)

### LINKED ISSUE #38: Not possible to specify sqlite db location
It seems like it is impossible to specify the location for the sqlite database. It would be nice if one of those or both could work:

- `kine --endpoint /var/lib/kine/state.db?more=rwc&_journal=WAL&cache=shared`
- `kine --endpoint file:///var/lib/kine/state.db?more=rwc&_journal=WAL&cache=shared`

ncopa:
Make it possible to set filepath to sqlite database via --endpoint
argument.

Fixes #38

### DIFF
--- a/pkg/endpoint/endpoint.go
+++ b/pkg/endpoint/endpoint.go
@@ -142,7 +142,7 @@ func ParseStorageEndpoint(storageEndpoint string) (string, string) {
 	network, address := networkAndAddress(storageEndpoint)
 	switch network {
 	case "":
-		return SQLiteBackend, ""
+		return SQLiteBackend, storageEndpoint
 	case "http":
 		fallthrough
 	case "https":


cjellick:
Thanks. Curious: how are you using kine? Is this just meant to make its way into k3s or are you using kine in another project. If so, could you share?

galal-hussein:
@ncopa this is already possible using sqlite://<path-on-disk> to enable sqlite backend with custom path

ncopa:
> Thanks. Curious: how are you using kine? Is this just meant to make its way into k3s or are you using kine in another project. If so, could you share?

experimenting with using kine with upsteam k8s.

> this is already possible using sqlite:// to enable sqlite backend with custom path

Oh, indeed. That was not evident while reading the code. Thanks alot!

Might be an idea to exit with error when invalid/unsupported arg is passed to endpoint rather than silently ignore it.

Issue #26, opened 2020-02-03
Publish official kine images


ibuildthecloud:
We should hook up CI on this repo and push rancher/kine images so people can use them as a sidecar for k8s apiservers.

simonferquel:
I manually built a kine image and used it as a sidecar for an aggregated API server. It works very well, and I really like this sidecar approach.
Actually all kubernetes distros putting the kube API server as a static kubelet manifest could also benefit from this, especially for single-node scenarios.

ibuildthecloud:
I merged #27, I'll leave this issue open until we actually start publishing the images.  We have drone infrastructure that does multi arch so it's not too hard to setup, somebody just need to create the drone.yaml.  I decent example is here https://github.com/rancher/go-skel/blob/master/.drone.yml  Just delete the steps that refer to dapper and only do the image build/manifest.

@ibrokethecloud Maybe you'd be interested in helping here?

ShadowJonathan:
@ibuildthecloud unless you use `exec`/`ssh` runners, you can't (yet) build OCI images inside docker without requiring privileged builds, and then using `kind` to allow the pipeline action to use the docker socket to build the images.

Maybe some images to look out for: https://github.com/containers/libpod/issues/4056 and https://github.com/containers/libpod/issues/4131, podman is trying to solve this issue by providing a rootless unprivileged way of building and running containers inside containers.

PR #41, opened 2020-07-03
Publish official kine images: Create .drone.yml

toonsevrin:
Public images of Kine would be very useful. As #26 has gone stale, I've taken the liberty to follow up on @ibuildthecloud 's request to add the .drone.yml.

I have absolutely zero experience with drone deployments so just copy pasted that file, edits by maintainers are thus allowed :smile:


toonsevrin:
I've got no idea why this pr messed up...

PR #45, opened 2020-07-12
Add Oracle backend support

stevefan1999-personal:
Well, let's hope that the "generic" interface is really generic enough for Oracle, one of the most egregious database system to date.

I've created this PR for...well you know Oracle offered free autonomous database and I don't want to miss a single opportunity to suck off those bastards. This also means this patch is potentially only for 19c and upwards.

But TBH I think using an ORM is a much better choice if you really want to generalize these cliche database difference. I suggest using GORM, and this could be another PR if I would like to work out the internal structure of this project.

Oh, If this gets patched I will keep on adding it to k3s too. That's actually my ultimate goal.


stevefan1999-personal:
I knew I had it coming. The nightmares of dealing with PL/SQL back in the days...are coming back to haunt me here.

On a bright side this thing is kind of working. I was able to get a connection to the autonomous database, but just as I said I ran into SQL issues. 

One of the most significant thing here is that you need to replace `LIMIT %d` to `FETCH FIRST %d ROWS ONLY`, so I will make it a `limit(count int)` suffix format function.

And another one is that Oracle Database...doesn't really have a built-in boolean types. While the statement did found out the token "true" to be a BOOLEAN, it was expecting a NUMBER instead, so it is not coerced by default LUL.

I also want to mention that `kv.deleted = (0 OR ?)` doesn't work in PL/SQL too, but it can be replaced as `WHERE kv.deleted = 0 OR kv.deleted = TO_NUMBER(?)`, well probably due to distributive law. The latter is more closer to ANSI SQL so can we use that instead?

Otherwise I think this is pretty okay.

pkg/drivers/oracle/oracle.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	dialect.Migrate(context.Background())
+	return logstructured.New(sqllog.New(dialect)), nil
+}
+
+func setup(db *sql.DB) error {
+	var str strings.Builder
+
+	for _, cmd := range procedureTemplate {
```
brandond, line 78: I am only lightly versed in oracle's SQL dialect, but why is it necessary to wrap all the queries in a templated procedure wrapper? Is that the only way to keep errors from messing up the connection state or something?

stevefan1999-personal:
This could be good. Sadly we don't have a test suite so I can validate my work

dereknola:
Is this PR still relevant?

PR #46, opened 2020-07-14
Add Gorm v2-based backend

stevefan1999-personal:
Now, as I mentioned in #45, I'm not quite happy about this project using the so called "generic" backend with handcrafted SQL that is not actually generic across all platforms. 

To truly alleviate yourself from all the craps differences on the database system we need to use ORM, and Gorm is the best out in the wild.

This also means we can delete ~300 lines of useless trash code and so there are less mental burdens and to focus more on implementing the rest of the features. I'm really hoping to see that in the future we can add more backends so that we can use Redis or Mongo😁.

btw Gorm has a compatibility interface to `sql.Sql` and transition from raw SQL to ORM should also be smoothy.

I've also added MSSQL support so that #43 can also be closed if this is merged for superseding.

But again, as I used a new public beta version of [Gorm](https://v2.gorm.io/), and so until V2 is GA this should not be merged. I use that for future proofing reason


stevefan1999-personal:
[Gorm V2 has been released](https://gorm.io/zh_CN/docs/changelog.html#v2-0-2020-08), so I think it is good to go now. This PR is open for merging

unixfox:
I'm interested into this PR because I want to use kine with an Oracle database.

yue9944882:
@stevefan1999-personal any plan to revive this thread?

stevefan1999-personal:
@yue9944882 Sure but I moved on and use their etcd backend instead, which i guess is better than using a shabby etcd wrapper that only covers the transaction extension

PR #50, opened 2020-07-29
Copy upstream behavior around setting maxIdle to 0

brandond:
As per https://github.com/rancher/kine/pull/37#issuecomment-665492471
setting this to 0 or allowing it to default to 0 imposes a severe
performance penalty.

### DIFF
--- a/pkg/drivers/generic/generic.go
+++ b/pkg/drivers/generic/generic.go
@@ -15,6 +15,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	defaultMaxIdleConns = 2 // copied from database/sql
+)
+
 var (
 	columns = "kv.id as theid, kv.name, kv.created, kv.deleted, kv.create_revision, kv.prev_revision, kv.lease, kv.value, kv.old_value"
 	revSQL  = `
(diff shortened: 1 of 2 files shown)


pkg/drivers/generic/generic.go
```diff
@@ -135,6 +135,10 @@ func (d *Generic) Migrate(ctx context.Context) {
 }
 
 func configureConnectionPooling(connPoolConfig ConnectionPoolConfig, db *sql.DB) {
+	if connPoolConfig.MaxIdle < 2 {
```
briandowns, line 138: Would it be possible to make this number a const with a descriptive name that might include the unit of measure, ie: `maxIdeSeconds = 2` or something appropriate?

brandond: I think that I actually see a better way to handle this, inspired by upstream:
https://github.com/golang/go/blob/release-branch.go1.13/src/database/sql/sql.go#L427

briandowns: Still requires the reader to have seen it first to reference back or know to check up there. This is considered a "magic number" and generally should be avoided.

PR #52, opened 2020-08-03
Fix compilation errors with NOCGO

brandond:
This is related to #53 - was turned up by enabling CI builds. 

### DIFF
--- a/pkg/drivers/dqlite/no_dqlite.go
+++ b/pkg/drivers/dqlite/no_dqlite.go
@@ -4,12 +4,12 @@ package dqlite
 
 import (
 	"context"
-	"fmt"
+	"errors"
 
 	"github.com/rancher/kine/pkg/drivers/generic"
 	"github.com/rancher/kine/pkg/server"
 )
 
 func New(ctx context.Context, datasourceName string, connPoolConfig generic.ConnectionPoolConfig) (server.Backend, error) {
-	return nil, fmt.Errorf("dqlite is not support, compile with \"-tags dqlite\"")
+	return nil, errors.New(`this binary is built without dqlite support, compile with "-tags dqlite"`)
 }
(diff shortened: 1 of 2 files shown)


pkg/drivers/dqlite/no_dqlite.go
```diff
@@ -11,5 +11,5 @@ import (
 )
 
 func New(ctx context.Context, datasourceName string, connPoolConfig generic.ConnectionPoolConfig) (server.Backend, error) {
-	return nil, fmt.Errorf("dqlite is not support, compile with \"-tags dqlite\"")
+	return nil, fmt.Errorf("this binary is built without dqlite support, compile with \"-tags dqlite\"")
```
briandowns, line 14: Since there's no value being passed in for use in the string formatting, `errors.New` is preferred.

brandond: Sure, I can clean that up while I'm here. I was mostly just irked by the poor grammar.

pkg/drivers/dqlite/no_dqlite.go
```diff
@@ -4,12 +4,12 @@ package dqlite
...
-	"fmt"
+	"errors"
 
 	"github.com/rancher/kine/pkg/drivers/generic"
 	"github.com/rancher/kine/pkg/server"
 )
 
 func New(ctx context.Context, datasourceName string, connPoolConfig generic.ConnectionPoolConfig) (server.Backend, error) {
-	return nil, fmt.Errorf("dqlite is not support, compile with \"-tags dqlite\"")
+	return nil, errors.New("this binary is built without dqlite support, compile with \"-tags dqlite\"")
```
briandowns, line 14:
```suggestion
	return nil, errors.New(`this binary is built without dqlite support, compile with "-tags dqlite"`)
```

brandond: The convention for this codebase seems to be to use double quotes instead of backticks for unformatted and/or wrapped error messages - should I change the rest of them as well?

briandowns: To follow go standards, yes, however the conventions of the code base should take precedence.  I would say it's worthwhile to do it in a separate PR and ignore this change request comment for now.

Issue #61, opened 2020-11-10
Investigate unexpected resource disappearance


brandond:
We have observed an instance of a single Kubernetes resource unexpectedly disappearing from the apiserver. Due to lack of apiserver audit logs it is unclear if it was removed at the Kubernetes level by a client operation, or if it was somehow incorrectly deleted by compaction code. K3s logs do not show any errors during the period in which we suspect the object was removed.

The backend datastore was a two-node AWS RDS Aurora MySQL cluster running version `5.7.mysql_aurora.2.09.0`. K3s version was `k3s v1.18.10+k3s2`.

We need to do additional testing of multi-node database clusters to ensure that all operations are safe when multiple kine instances are connected to a database cluster. In initial review of the code does not turn up any remaining obviously unsafe operations. There were some questions about transaction safety of inserts, but an initial cursory review of the go-mysql code suggests that this should be safe to use the way we're using it.

gz#13317

MerlinDMC:
Interesting to see this popping up here as I'm in the process of trying to find the cause for our resource disappearance.
We however run Kubernetes 1.18 against a normal Etcd cluster and observed secrets, deployments and statefulsets getting terminated in the events stream during higher load on the API server and Etcd. (we replaced worker nodes and the scheduler started reassigning workloads)

That leaves me slightly worried now as that probably rules out our slightly older Etcd version and it might be in the API server 😞 

And while I'm here - thx for https://github.com/etcd-io/etcd/pull/11776 as that is currently blocking my test environment 😉

brandond:
@MerlinDMC if you've seen something similar on clusters running stock etcd that makes me feel slightly more optimistic that there isn't some weird issue with kine that's disappearing our resources out from underneath us.

MerlinDMC:
yes - but I was just about to try kine as a replacement to get rid of the problem 🤣

brandond:
Can you share any more details on what you were seeing? Was the apiserver actually deleting things from etcd? How were you monitoring the issue?

MerlinDMC:
I was monitoring status of the deployments via `kubectl get --all-namespaces events -w`. At some point during the rollout of our services to new nodes (roughly 1000 pods) I did see terminations for a lot of deployments in the event stream.

I'm not sure yet if the apiserver did issue those deletes or if it did read keys from Etcd and didn't get all the data back. No person however did issue deletes.

I'm trying to get a test scenario up that has the same application load so we can try to forcefully get into the same issue again and collect more data.

[...]

PR #62, opened 2020-11-13
Fix bad check in compact query

brandond:
* Fix issue where compact_rev_key was protected from deletion, instead
  of the row it points to. Honestly we should have stored the
  compact_rev marker in the value instead of the prev_revision
  column. (#61)

Test:
1. Run kine for k3s with external DB: `k3s --debug server --datastore=mysql://X`
1. Wait for compaction message to appear in log output - something like `level=debug msg="COMPACT deleted 0 rows from 3 revisions in 2.484763ms - compacted to 3/1745"`
1. Stop k3s+kine
1. Run query: `SELECT id, name FROM kine WHERE name = "/registry/health";`
1. Run query: `UPDATE kine SET prev_revision = 1 WHERE name = 'compact_rev_key';` - use the ID from the previous step instead of `1` if it was different.
1. Start k3s; wait for compaction message to appear in log output
1. Run query: `SELECT id, name FROM kine WHERE name = "/registry/health"`
1. Confirm that /registry/health entry still present with the same ID

### DIFF
--- a/pkg/drivers/mysql/mysql.go
+++ b/pkg/drivers/mysql/mysql.go
@@ -74,6 +74,7 @@ func New(ctx context.Context, dataSourceName string, tlsInfo tls.Config, connPoo
 			SELECT kp.prev_revision AS id
 			FROM kine AS kp
 			WHERE
+				kp.name != 'compact_rev_key' AND
 				kp.prev_revision != 0 AND
 				kp.id <= ?
 			UNION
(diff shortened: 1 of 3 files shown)


ibuildthecloud:
LGTM

briandowns (approved):

pkg/drivers/mysql/mysql.go
```diff
@@ -71,21 +71,26 @@ func New(ctx context.Context, dataSourceName string, tlsInfo tls.Config, connPoo
...
+				ORDER BY kp.id ASC
+				LIMIT 1000
+			) UNION ALL (
+				SELECT kd.id AS id
+				FROM kine AS kd
+				WHERE
+					kd.deleted != 0 AND
+					kd.id <= ?
+				ORDER BY kd.id ASC
+				LIMIT 1000
```
cjellick, line 90: for anyone else watching, brad's going to drop this limit change to minimize the delta for the fix

galal-hussein (approved):
LGTM

brandond:
Can tackle query optimizations later, for now how about we just stop deleting things we shouldn't. Patch is updated.

cjellick (approved):

briandowns (approved):

PR #64, opened 2020-11-19
Feat: Skips ddl stmt upon setting up the database

yue9944882:
our production environment doesn't allow the application executes DDL statement at application runtime, the CREATE/ALTER statements are required to be submitted/reviewed in other places. this pull allows developers to opt-out from the DDL commands

### DIFF
--- a/pkg/drivers/mysql/mysql.go
+++ b/pkg/drivers/mysql/mysql.go
@@ -40,7 +40,8 @@ var (
 		`CREATE INDEX kine_prev_revision_index ON kine (prev_revision)`,
 		`CREATE UNIQUE INDEX kine_name_prev_revision_uindex ON kine (name, prev_revision)`,
 	}
-	createDB = "CREATE DATABASE IF NOT EXISTS "
+	createDB         = "CREATE DATABASE IF NOT EXISTS "
+	checkSchemaExist = "SELECT 0 from kine"
 )
 
 func New(ctx context.Context, dataSourceName string, tlsInfo tls.Config, connPoolConfig generic.ConnectionPoolConfig) (server.Backend, error) {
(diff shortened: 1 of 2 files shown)


brandond:
This is an interesting request, I will have to talk it over with the team.

I would suggest that we only support this for postgres/mysql databases - so drop the change to the sqlite driver. I do not think it is reasonable to have someone externally manage the DDL for an embedded database engine.

yue9944882:
> Also, if we're not going to attempt to create the tables and indexes on startup, I think we need some fallback code to ensure their existence.

thanks, checking the table existence SGTM, plz let me know what's your team thought. will update as follow-up if that makes sense..

ibuildthecloud:
This idea was that all the statements should be "create if not exists" allowing them to be created beforehand. What I don't think I thought about is that you will be rejected because you don't have to create privileges even if the object exists and you don't need to create it.

If we can just check first if something exist before creating it, that's would be preferred.

brandond:
@ibuildthecloud I thought that the 'create table if not exists' that we're doing would be safe enough but I think you're right that these would probably result in permissions errors even if the table exists. For one of the engines we aren't doing if-not-exists; we're just running the create and ignoring 'already exists' errors.

Explicitly checking for each table and index before creating it seems like a fair bit more code to maintain, but ultimately I think it'll be easier to support than having a flag to completely externalize DDL maintenance - since we really want to validate the existing schema if we're not blindly ensuring that it exists.

yue9944882:
@brandond @ibuildthecloud i updated the pull to probe the existence of schema before executing create table statements, this will be more reasonable than simply providing a skip option.

dereknola:
Is this still relevant?

PR #70, opened 2020-12-11
Explicitly use ascii charset for the name column 

yue9944882:
> MySQL [xxx]> CREATE UNIQUE INDEX kine_name_prev_revision_uindex ON kine (name, prev_revision);
ERROR 1071 (42000): Specified key was too long; max key length is 767 bytes

creating index will failing on mysql prior to 5.6 version, explicitly using ascii charset will workaround the failure

https://stackoverflow.com/a/1814594

### DIFF
--- a/pkg/drivers/mysql/mysql.go
+++ b/pkg/drivers/mysql/mysql.go
@@ -24,7 +24,7 @@ var (
 		`CREATE TABLE IF NOT EXISTS kine
 			(
 				id INTEGER AUTO_INCREMENT,
-				name VARCHAR(630),
+				name VARCHAR(630) CHARACTER SET ascii,
 				created INTEGER,
 				deleted INTEGER,
 				create_revision INTEGER,


pkg/drivers/mysql/mysql.go
```diff
@@ -24,7 +24,7 @@ var (
 		`CREATE TABLE IF NOT EXISTS kine
 			(
 				id INTEGER AUTO_INCREMENT,
-				name VARCHAR(630),
+				name VARCHAR(630) character set ascii,
```
brandond, line 27:
```suggestion
				name VARCHAR(630) CHARACTER SET ascii,
```

brandond (changes requested):

pkg/drivers/mysql/mysql.go
yue9944882: @brandond thanks for the prompt review, updated now

brandond (approved):

dereknola (approved):

PR #81, opened 2021-03-26
Add synthetic perf tests with K3s in docker

brandond:
Adds E2E perf tests against commonly used database engines:
```
    mysql 8.0
    mysql 5.7
    mysql 5.6
    mariadb 10.5
    mariadb 10.4
    mariadb 10.3
    mariadb 10.2
    postgres 13.2
    postgres 12.6
    postgres 11.11
    postgres 10.16
    postgres 9.6
    postgres 9.5
    cockroachdb v20.2
    cockroachdb v20.1
    cockroachdb v19.2
```
Tests are performed using K3s in Docker with only the apiserver enabled, and a script that performs random conflicting create/update/list/delete operations to ConfigMaps. At the end of each run, performance statistics (average + a histogram) are output for comparison across runs and engines. These statistics are retrieved from the apiserver's etcd request latency metrics.


hack/histogram.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def main(type, *args, **kwargs):
+    global v1
+
+    try:
+        config.load_kube_config()
+        version = client.VersionApi().get_code()
+        logging.info(f"Connected to {Configuration._default.host} - {version.git_version}")
+    except Exception as e:
```
briandowns, line 26: Is there any specific behavior that might be associated with a particular exception we would want to catch or do we not care and just bail on any?

brandond: This is essentially a baseline configuration + connection test. Since we're both loading the kubeconfig and making a request to the server to get the version, any basic error (bad config, can't connect to server, bad creds, etc) is handled here.

briandowns (approved):

erikwilson (approved):

Issue #84, opened 2021-05-17
Retries in generic driver


ktsakalozos:
Thank you for kine! I have a couple of questions on the retry policy used.

1. I am trying to understand the retry time interval set in [1]. Trying the back-off strategy used (`backoff.Linear(100 + time.Millisecond)`) it seems too aggressive to me [2], but I do not have any data to prove this. It seems we will retry 20 times the same operation within 171 milliseconds. The `backoff.Linear()` takes a Duration as parameter and we currently pass in 1 millisecond and 100 nanoseconds. Aren't the 100 nano seconds shadowed by the 1 millisecond? I suspect the intention was to have something like `100 *  time.Millisecond`.

1. We retry only the `execute()` call [3]. What about the `query()` [4] and `queryRow()` [5] calls? Shouldn't these be retried as well?

Thanks

[1] https://github.com/k3s-io/kine/blob/master/pkg/drivers/generic/generic.go#L256
[2] https://play.golang.org/p/pBxe-DUqhy-
[3] https://github.com/k3s-io/kine/blob/master/pkg/drivers/generic/generic.go#L250
[4] https://github.com/k3s-io/kine/blob/master/pkg/drivers/generic/generic.go#L240
[5] https://github.com/k3s-io/kine/blob/master/pkg/drivers/generic/generic.go#L245

PR #85, opened 2021-05-17
Retry every 50ms and include more operations

ktsakalozos:
Fixes: https://github.com/k3s-io/kine/issues/84

Signed-off-by: Konstantinos Tsakalozos <kos.tsakalozos@canonical.com>


brandond:
I think starting at 50ms is way too high - datastore operations usually complete in several milliseconds tops, so the intent is for this to basically function as a busy-wait in case we hit some weird transaction edge or locking issue - as opposed to a more traditional retry of a network operation.

ktsakalozos:
@brandond you are right a few miliseconds is the right scale [1]. It is hard to set a value that will optimally address the needs of all drivers in all of their setups. So I will not touch the waiting strategy in this PR, I will leave the `100 + time.milisecond` as is (although I do not understand it 100%).

On this PR I am only extending the retries to cover not only the execute call but the query as well.

[1] https://etcd.io/docs/v3.4/op-guide/performance/

ktsakalozos:
I will probably followup with a PR to make the waiting strategy somewhat configurable from within the driver used.

PR #86, opened 2021-05-21
Feat: automatically log slow sql executions

zqzten:
This PR adds feature to automatically log slow sql.

It's common that sometimes network or database problems occur in production which cause slow sql executions and lead to apiserver errors. It would be handy to automatically log some sql execution details when we meet a slow sql (only log slow sqls not to make the INFO log too verbose) for troubleshooting afterwards. It's also helpful for detecting real slow sql problems in real production use cases.

### DIFF
--- a/pkg/drivers/generic/generic.go
+++ b/pkg/drivers/generic/generic.go
@@ -239,11 +239,15 @@ func Open(ctx context.Context, driverName, dataSourceName string, connPoolConfig
 
 func (d *Generic) query(ctx context.Context, sql string, args ...interface{}) (*sql.Rows, error) {
 	logrus.Tracef("QUERY %v : %s", args, Stripped(sql))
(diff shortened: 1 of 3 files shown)


brandond:
most SQL engines already have built-in options for doing this, for example: https://dev.mysql.com/doc/refman/5.7/en/slow-query-log.html - the only place I could see this being useful is perhaps with sqlite?

zqzten:
@brandond Thanks for the reply. This feature trace the whole SQL execution time which includes actual SQL execution time in database and also the connection and transporting time. We've met issues that a client took a long time executing a normal SQL (which did not cause a slow SQL in db), that's why we want to do it - this feature can also help to find out other possible problems along with the database log.

zqzten:
Replaced by a more general metrics implementation #121

PR #92, opened 2021-06-29
Ensure compaction runs during startup

brandond:
* Ensure that initial compaction runs during Start. Deferring this until
Watch creates a race condition wherein we can start answering
requests before compact_rev_key is created, which will cause SQL errors
as seen in k3s-io/k3s#3523.

* Fix client.Delete; the current implementation does not actually work
because delete is only implemented within transactions.

* Fix CI tests by pinning the library we're using to output perf stats

### DIFF
--- a/pkg/logstructured/sqllog/sql.go
+++ b/pkg/logstructured/sqllog/sql.go
@@ -53,9 +53,9 @@ type Dialect interface {
 	BeginTx(ctx context.Context, opts *sql.TxOptions) (*generic.Tx, error)
 }
 
-func (s *SQLLog) Start(ctx context.Context) (err error) {
+func (s *SQLLog) Start(ctx context.Context) error {
 	s.ctx = ctx
-	return
+	return s.compactStart(s.ctx)
 }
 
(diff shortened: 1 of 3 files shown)


galal-hussein (approved):
LGTM

pkg/logstructured/sqllog/sql.go
```diff
@@ -53,9 +53,12 @@ type Dialect interface {
 	BeginTx(ctx context.Context, opts *sql.TxOptions) (*generic.Tx, error)
 }
 
-func (s *SQLLog) Start(ctx context.Context) (err error) {
+func (s *SQLLog) Start(ctx context.Context) error {
 	s.ctx = ctx
-	return
+	if err := s.compactStart(s.ctx); err != nil {
```
briandowns, line 58:
```suggestion
	return s.compactStart(s.ctx)
```

brandond: I wish gofmt would carp about this

briandowns (approved):
LGTM

PR #101, opened 2021-08-18
Configure GRPC keepalives

brandond:
Related to:
* https://github.com/k3s-io/k3s/issues/3881


dweomer (approved):
LGTM but I am hoping that the addition of the etcd/server package does not introduce too much bloat into k3s!

brandond:
@dweomer on the k3s side we already import that package for the embedded etcd server, so I don't think it should have much impact. It does make the standalone kine binary about 5mb bigger (13M -> 18M) though.

Issue #30, opened 2020-03-29
Implement etcd version endpoint


angelbarrera92:
Hi Rancher team!

I am testing this etcdshin using kubeadm. It works!!!

But every time I use it with kubeadm i have to use the following flag: `--ignore-preflight-errors=ExternalEtcdVersion` because kine seems like does not implement this endpoint.

Is this intended? What version of etcd is implemented? 3.X? I can PR if it should be implemented.

ibuildthecloud:
We only implemented the apis needed by k8s. I'd make the argument that if kubeadm needs the API we should add it. I'd say this is etcd 3.4.  would you be willing to create a PR for this. It should be straight forward because it's a hard coded response.

PR #104, opened 2021-08-27
Add support for etcd version, memberList, and status queries

brandond:
In upstream etcd, the version endpoint is implemented as a simple JSON handler, muxed with the GRPC server on the client port. This PR implements the required muxing for both plaintext and TLS listeners, using an approach inspired by upstream etcd. Kine now sends a hardcoded etcd version response that should satisfy clients such as kubeadm's [preflight checks](https://github.com/kubernetes/kubernetes/blob/release-1.21/cmd/kubeadm/app/preflight/checks.go#L694) (fixing #30).

This PR also splits the server code into different files per GRPC service, and adds a Cluster service that only supports GetMembers. The service returns a fixed member list that uses the GRPC ":authority" metadata (analogous to the HTTP Host header) to set the returned peer and client URLs. This should be sufficient to satisfy clients such as [KCP](https://github.com/kcp-dev/kcp) that enumerate the etcd member list as an informational [startup check](https://github.com/kcp-dev/kcp/blob/main/cmd/kcp/kcp.go#L105).

Third, this PR adds a Maintenance service that only supports Status, populating the DbSize field if the driver has a query to report this data. Currently sqlite, mysql, and postgres all support this; cockroachdb does not. This field is periodically collected by the Kubernetes apiserver and copied into the etcd_db_total_size_in_bytes metric.

### DIFF
--- a/pkg/endpoint/endpoint.go
+++ b/pkg/endpoint/endpoint.go
@@ -16,6 +16,7 @@ import (
 	"github.com/k3s-io/kine/pkg/tls"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"github.com/soheilhy/cmux"
 	"go.etcd.io/etcd/server/v3/embed"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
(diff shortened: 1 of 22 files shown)


pkg/endpoint/endpoint.go
```diff
@@ -65,44 +66,107 @@ func Listen(ctx context.Context, config Config) (ETCDConfig, error) {
...
 
-func createListener(listen string) (ret net.Listener, rerr error) {
-	network, address := networkAndAddress(listen)
+// endpointURL returns a URI string suitable for use as a local etcd endpoint.
+// For TCP sockets, it is assumed that the port can be reached via the loopback address.
+func endpointURL(config Config, listener net.Listener) string {
+	scheme := endpointScheme(config)
+	address := listener.Addr().String()
+	if scheme != "unix" {
+		address = "127.0.0.1" + address[strings.LastIndex(address, ":"):]
```
dweomer, line 135: Consider leveraging https://pkg.go.dev/net#SplitHostPort

pkg/endpoint/http.go
```diff
@@ -0,0 +1,66 @@
...
+	"log"
+	"net/http"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+)
+
+var (
+	etcdVersion = []byte(`{"etcdserver":"3.5.0","etcdcluster":"3.5.0"}`)
```
dweomer, line 13: Would be slick if this version could be lifted from etcd server code somewhere but then it would likely have to be injected at build time :man_shrugging:

dweomer (approved):

pkg/endpoint/http.go
brandond: Yeah, I considered getting more clever with it but ended up just hardcoding it as suggested at https://github.com/k3s-io/kine/issues/30#issuecomment-622977646

pkg/endpoint/endpoint.go
brandond: thanks! I was looking for something a little more lightweight than a full url parser but didn't come across this.

galal-hussein (approved):

dweomer (approved):

PR #115, opened 2022-03-03
Add initial JetStream driver to kine

boxboatmatt:
This PR proposes to add [NATS JetStream](https://docs.nats.io/nats-concepts/jetstream) as a backend option for kine. NATS adoption on the edge is rapidly growing and backing k3s to NATS is a great fit in environments where NATS is already being employed.

JetStream, which enables new functionalities and higher qualities of service on top of the base 'Core NATS' functionalities and qualities of service, is built-in to nats-server and you only need 1 (or 3 or 5 if you want fault-tolerance against 1 or 2 simultaneous NATS server failures) of your NATS server(s) to be JetStream enabled for it to be available to all the client applications.


brandond:
Thanks for the contribution, this is very interesting! I see there are still some bits commented out, and a few things that are waiting on upstream PRs. Did you want to get those worked out before merging?

pkg/endpoint/endpoint.go
```diff
@@ -233,6 +235,8 @@ func getKineStorageBackend(ctx context.Context, driver, dsn string, cfg Config)
 		backend, err = pgsql.New(ctx, dsn, cfg.BackendTLSConfig, cfg.ConnectionPoolConfig)
 	case MySQLBackend:
 		backend, err = mysql.New(ctx, dsn, cfg.BackendTLSConfig, cfg.ConnectionPoolConfig)
+	case JetStreamBackend:
+		backend, err = jetstream.New(ctx, dsn)
```
brandond, line 239: Does Jetstream support TLS or connection pooling?

scripts/test-helpers
```diff
@@ -159,7 +159,12 @@ fetch-kubeconfig() {(
     local num=${1:-1}
     local name=$(cat $TEST_DIR/servers/$num/metadata/name)
     local url=$(cat $TEST_DIR/servers/$num/metadata/url)
-    docker cp $name:/etc/rancher/k3s/k3s.yaml - 2>/dev/null | tar -xO 2>/dev/null | sed -e "s|https://127.0.0.1:6443|$url|g" >$TEST_DIR/servers/$num/kubeconfig.yaml
+    # check for macos and do not replace k8s api url
+    if [[ "$OSTYPE" =~ ^darwin ]]; then
+      docker cp $name:/etc/rancher/k3s/k3s.yaml - 2>/dev/null | tar -xO 2>/dev/null  >$TEST_DIR/servers/$num/kubeconfig.yaml
```
brandond, line 164: I'm curious about this - what is different about OS X that the server URL doesn't need to be updated in order to connect to the port exposed by the container?

pkg/drivers/jetstream/jetstream.go
```diff
@@ -0,0 +1,900 @@
...
+)
+
+const (
+	kineBucket             = "kine"
+	revHistory             = 12
+	slowMethodMilliseconds = 500
+)
+
+var (
+	toplevelKeyMatch = regexp.MustCompile(`/registry/(.*)/.*`)
```
brandond, line 25: K3s stores cluster bootstrap data outside the Kubernetes keyspace under the `/bootstrap/` prefix - does this need any special consideration?
https://github.com/k3s-io/k3s/blob/04c39ba621d1ca30086a96694d74fe23f2337228/pkg/cluster/encrypt.go#L21-L23

scripts/test-helpers
dweomer: this does seem like a superfluous conditional. what if youre k3s installation is kind-like aka k3d?

boxboatmatt: There may be a better way to address the issue for running these tests locally on OS X.... The issue on OS X is that docker machine is a VM so the networking gets weird... I added 2 changes here one was to bind 6443 to 6443 on localhost rather than a randomly assigned port. And then not to replace the 127.0.0.1 with what is getting pulled out here:

```
    local ip=$(docker container inspect --format '{{.NetworkSettings.IPAddress}}' $name | tee $TEST_DIR/kine/$count/metadata/ip)
    local port=$(docker container inspect --format '{{range $k, $v := .NetworkSettings.Ports}}{{printf "%s\n" $k}}{{end}}' $name | head -n 1 | cut -d/ -f1 | tee $TEST_DIR/kine/$count/metadata/port)
    local url=$(echo "http://$ip:$port" | tee $TEST_DIR/kine/$count/metadata/url)
```

Without those changes you will get something like this locally on a test run:
```
[nats-2.7.3] Started k3s-server-1-KgHGdM @ https://172.17.0.4:6443
[nats-2.7.3] https://172.17.0.4:6443
[nats-2.7.3] Waiting for kubeconfig to become available...
[nats-2.7.3] https://172.17.0.4:6443
[nats-2.7.3] Running configmap load generation script
[nats-2.7.3] [33353]	W Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x10bb856d0>: Failed to establish a new connection: [Errno 60] Operation timed out')': /version/
[nats-2.7.3] [33350]	W Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x10e625760>: Failed to establish a new connection: [Errno 60] Operation timed out')': /version/
[nats-2.7.3] [33352]	W Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x109f6c5b0>: Failed to establish a new connection: [Errno 60] Operation timed out')': /version/
[nats-2.7.3] [33351]	W Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x111d60070>: Failed to establish a new connection: [Errno 60] Opera
```

The local python k8s client can't connect to url so it fails... But thinking about this now perhaps we can leave the randomly assigned port and connect on `127.0.0.1:<random-port>` but that still might need some special OS X handling.

dweomer: yeah, kind has a port-forwarding config i've had to leverage on os-x and k3s will. iirc, expose a load-balancer on a host-local interface. if we're using docker directly we have to wing it. maybe this is fine for now?

dweomer:
@boxboatmatt thank you for this contribution! Based on my light reading, I have some concerns that you can likely speak to:
- as @brandond asked [here](https://github.com/k3s-io/kine/pull/115#discussion_r819026155), can I haz TLS? 
- keyspace seems to be encoded on the wire. will we need a decoder on the nats backend or is this builtin?
- performance hit for keyspace encoding
  - also i see the two-way hash and it's implied provenance ... are we to assume that a search through potential collision space has been performed exhaustively?

pkg/endpoint/endpoint.go
boxboatmatt: TLS yes... connection pooling no

I should definitely expose the TLSConfig and revisit the connection logic

boxboatmatt: For reference:
https://github.com/nats-io/nats.go#tls

boxboatmatt:
> @boxboatmatt thank you for this contribution! Based on my light reading, I have some concerns that you can likely speak to:
> 
> * as @brandond asked [here](https://github.com/k3s-io/kine/pull/115#discussion_r819026155), I can haz TLS?

Yup TLS support is definitely there - I just need to add it... My intention was to revisit the connection but I haven't done that yet :)


> * keyspace seems to be encoded on the wire. will we need a decoder on the nats backend or is this builtin?
> * performance hit for keyspace encoding
>   
>   * ~also i see the two-way hash and it's implied provenance ... are we to assume that a search through potential collision space has been performed exhaustively? (a somewhat silly question to two-way hash/encoding but sometimes my brain goes places)~

NATS streams reserves `.` using so the key space has to be encoded to work properly with etcd keys which reserves `/` and can contain `.` within the keys. I am sure there is some performance hit there but if you look at the load-map numbers in the build it is still very fast.

pkg/drivers/jetstream/jetstream.go
boxboatmatt: Ah yes... I should probably just add handling for it just in case - I don't think as written any problems would occur but it won't hurt to handle that `/bootstrap/` case

boxboatmatt:
> Thanks for the contribution, this is very interesting! I see there are still some bits commented out, and a few things that are waiting on upstream PRs. Did you want to get those worked out before merging?

The nats.go [PR](https://github.com/nats-io/nats.go/pull/904) doesn't _have_ to get merged it would just make the client a little cleaner.

I should however remove the commented out pieces that are still lingering. I'll address those in addition to other comments

brandond:
> NATS JetStream reserves `.` so the key space has to be encoded to work properly with etcd keys which reserves `/` and can contain `.` within the keys. I am sure there is some performance hit there but if you look at the load-map test numbers in the build it is still very fast.

I'm curious, would it work to essentially just `tr './' '/.'` to swap the two reserved key components? I haven't actually thought this through since I'm not sure what meaning the periods have to JetStream. As of etcdv3 forward slashes actually don't have any meaning to etcd; keys are just arbitrary byte arrays.

pkg/endpoint/endpoint.go
jnmoyne: There is no need for connection pooling in JetStream many operations are multiplexed in parallel over the single connection (which can be TLS or un-encrypted) to (one of the) nats-server(s).

jnmoyne:
> I'm curious, would it work to essentially just transliterate periods in the key string to some other non-reserved character? I > haven't actually thought this through since I'm not sure what meaning the periods have to JetStream.

Dots separate 'tokens' in a NATS subject such that you can use wildcards (`*` and `>`) to match any value for one or more tokens in a subject when subscribing. You could indeed translate the periods in your key to some other non-reserved character (essentially any characters other than `.`,`*` and `>` are ok to use).

Here it's the key is just being base58 encoded which is going to be pretty fast encode/decode.

pkg/endpoint/endpoint.go
boxboatmatt: TLS Support added

pkg/drivers/jetstream/jetstream.go
boxboatmatt: Modified regex to allow for other TL folders like `/bootstrap`

pkg/drivers/jetstream/jetstream.go
```diff
@@ -90,6 +128,92 @@ func New(ctx context.Context, connection string) (server.Backend, error) {
...
+	if ctxOpt, ok := queryMap["context"]; ok {
+		logrus.Infof("loading nats context=%s", ctxOpt[0])
+		natsContext, err := natscontext.New(ctxOpt[0], true)
+		if err != nil {
+			return "", "", nil, err
+		}
+		// command line options provided to kine will override the file
+		// https://github.com/nats-io/jsm.go/blob/v0.0.29/natscontext/context.go#L257
+		// allows for user, creds, nke, token, certifcate, ca, inboxprefix from the context.json
+		natsClientOpts, err := natsContext.NATSOptions(o...)
```
caleblloyd, line 206: NATS Context also defines [ServerURL](https://github.com/nats-io/jsm.go/blob/48ea938560dce28b1170ed7d909489454de67792/natscontext/context.go#L451), but that seems to be ignored here...

I think that if a context is used, then the URLs should _only_ come from the context - throw an error if `host` or `port` is set

So a context url would be `nats://?context=/etc/nats.json`

pkg/drivers/jetstream/jetstream.go
```diff
@@ -41,20 +46,53 @@ type JSValue struct {
...
+	if err != nil {
+		return nil, err
 	}
-	logrus.Infof("connecting to %s", connection)
 
-	conn, err := nats.Connect(connection)
+	logrus.Infof("using bucket: %s", b)
+	logrus.Infof("connecting to %s", c)
+
+	nopts = append(nopts, nats.Name("k3s-server using bucket: "+b))
```
brandond, line 93:
```suggestion
	nopts = append(nopts, nats.Name("kine using bucket: "+b))
```

pkg/drivers/jetstream/jetstream.go
boxboatmatt: Seems reasonable to me

boxboatmatt: Updated to force this scheme

pkg/drivers/jetstream/jetstream.go
```diff
@@ -47,11 +47,12 @@ type JSValue struct {
...
 
 // New get the JetStream Backend, establish connection to NATS JetStream. At the moment nats.go does not have
-// connection string support so kine will use nats://(token|username:password)hostname:port?bucket=bucketName&context=nats-context`
+// connection string support so kine will use nats://(token|username:password)hostname:port?bucket=bucketName&contextFile=nats-context`.
+// If contextFile is provided then do not provide a hostname:port in the endpoint URL
 //
 // bucket: specifies the bucket on the nats server for all of the k3s values for this cluster (optional)
 //
-// context: specifies the nats context to load from ~/.config/nats/context/ e.g. nats-context for ~/.config/nats/context/nats-context.json
+// contextFile: specifies the nats context to load from ~/.config/nats/context/ e.g. nats-context for ~/.config/nats/context/nats-context.json
```
brandond, line 55: This is kind of confusing, since it's not strictly the file, but the file to be loaded is derived from the name provided here. I think it should either remain just `contex` and automatically derive the filename of the context to load, or it should be `contextFile` and require the full path+name+extension of the context to be loaded.

boxboatmatt: You're right... I missed updating this comment, the intent was to only support `contextFile=/some/path/context.json` and that is how it is currently implemented

pkg/drivers/jetstream/kv/kv.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+func (e *EncodedKV) newWatcher(w nats.KeyWatcher) nats.KeyWatcher {
+	watch := &watcher{
+		watcher:    w,
+		keyCodec:   e.keyCodec,
+		valueCodec: e.valueCodec,
+		updates:    make(chan nats.KeyValueEntry, 32)}
+
+	// waiting on https://github.com/nats-io/nats.go/pull/904 to refactor
```
caleblloyd, line 135: https://github.com/nats-io/nats.go/pull/904 has been merged; this can now be refactored

caleblloyd, line 135: https://github.com/nats-io/nats.go/pull/904 has been merged; this can now be refactored

pkg/drivers/jetstream/jetstream.go
```diff
@@ -0,0 +1,992 @@
...
+	"github.com/k3s-io/kine/pkg/server"
+	"github.com/k3s-io/kine/pkg/tls"
+	"github.com/nats-io/jsm.go/natscontext"
+	"github.com/nats-io/nats.go"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	kineBucket             = "kine"
+	revHistory             = 12
```
caleblloyd, line 24: It would be nice if `bucketRevisionHistory` and `slowMethodMilliseconds` were added to the DSN so that users can configure these

caleblloyd, line 24: It would be nice if `bucketRevisionHistory` and `slowMethodMilliseconds` were added to the DSN so that users can configure these

boxboatmatt: Added `slowMethodMs` and `revHistory` as optional query parameters

pkg/drivers/jetstream/jetstream.go
```diff
@@ -126,62 +138,85 @@ func New(ctx context.Context, connection string, tlsInfo tls.Config) (server.Bac
...
 	}
 
 	if b, ok := queryMap["bucket"]; ok {
-		bucketName = b[0]
+		jsConfig.bucket = b[0]
+	}
+
+	if r, ok := queryMap["slowMethodMs"]; ok {
+		if ms, err := strconv.Atoi(r[0]); err == nil {
+			jsConfig.slowMilliseconds = int64(ms)
```
brandond, line 173: nit:
```suggestion
		if ms, err := strconv.ParseInt(r[0], 10, 64); err == nil {
			jsConfig.slowMilliseconds = ms
```

pkg/drivers/jetstream/jetstream.go
```diff
@@ -0,0 +1,1017 @@
...
+	slowMilliseconds int64
+}
+
+type JetStream struct {
+	kvBucket         nats.KeyValue
+	kvBucketMutex    *sync.RWMutex
+	kvDirectoryMutex *sync.RWMutex
+	kvDirectoryMuxes map[string]*sync.RWMutex
+	jetStream        nats.JetStreamContext
+	slowMethod       int64
```
brandond, line 47: A more idiomatic way to specify this would be to use a `time.Duration` instead of an integer number of milliseconds; this would save you a bit of code on the parameter parsing side too.

brandond (changes requested):

pkg/drivers/jetstream/jetstream.go
boxboatmatt: I thought about doing that but then thought maybe that would provide too much flexibility... i.e. someone says a slow method is `1hr` but using a duration would probably make more sense

brandond (approved):

PR #118, opened 2022-04-26
Import etcd v3 consistently

cmurphy:
Fix the rpctypes module import to use the etcd/v3 package version. This
allows packages that include kine as a dependency to not have to include
the old etcd package.


brandond (approved):
Thanks!

PR #121, opened 2022-05-24
Introduce metrics

### LINKED ISSUE #119: metrics support
In production we want to collect some key metrics of kine for monitoring and alerting, such as the [DBStats](https://pkg.go.dev/database/sql#DBStats) of Go SQL backend used by kine, SQL operation times and errors etc. Would it be applicable that we add some prometheus metrics for them?

For library mode, we can add an injectable registerer in endpoint config and for stand-alone, we can setup our own metrics handler like etcd.

zqzten:
This PR introduce metrics to kine.

It'll expose below metrics:
* SQL Total & Errors: Counter metrics which hold the total number of SQL operations and failed SQL operations. This is helpful for quickly tracking and alerting local SQL errors which have a variant of causes (network, db, etc.) thus not easy to monitor.
* SQL Time: Histogram metric which keeps track of the duration of SQL operations. This duration is measured locally and can be used as a supplement to the SQL exec time measurement of db.
* Compact Errors: Counter metrics which hold the total number of compact errors. Helpful for alerting when continuous compact errors occur which can lead to uncontrollable growth of db size.
* [DBStats metrics](https://github.com/prometheus/client_golang/pull/866) of sql.DB used by kine.
* Process and Go runtime metrics when running stand-alone.

If kine is running stand-alone, we can use the `metrics-bind-address` flag to configure the metrics http server address or disable the server.
If kine is used as a library, we can set `MetricsRegisterer` of `endpoint.Config` to our own prometheus registerer to let kine register its metrics to it.

fixes #119

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,8 +5,10 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/k3s-io/kine/pkg/endpoint"
+	"github.com/k3s-io/kine/pkg/metrics"
 	"github.com/k3s-io/kine/pkg/version"
 	"github.com/rancher/wrangler/pkg/signals"
 	"github.com/sirupsen/logrus"
@@ -14,7 +16,8 @@ import (
 )
 
 var (
-	config endpoint.Config
+	config        endpoint.Config
+	metricsConfig metrics.Config
 )
 
 func main() {
@@ -76,6 +79,18 @@ func main() {
 			Usage:       "Key file for DB connection",
 			Destination: &config.BackendTLSConfig.KeyFile,
 		},
+		cli.StringFlag{
+			Name:        "metrics-bind-address",
+			Usage:       "The address the metric endpoint binds to. Default :8080, set 0 to disable metrics serving.",
+			Destination: &metricsConfig.ServerAddress,
+			Value:       ":8080",
+		},
+		cli.DurationFlag{
+			Name:        "slow-sql-threshold",
+			Usage:       "The duration which SQL executed longer than will be logged. Default 1s, set <= 0 to disable slow SQL log.",
+			Destination: &metrics.SlowSQLThreshold,
+			Value:       time.Second,
+		},
 		cli.BoolFlag{Name: "debug"},
 	}
 	app.Action = run
@@ -92,6 +107,9 @@ func run(c *cli.Context) error {
 		logrus.SetLevel(logrus.TraceLevel)
 	}
 	ctx := signals.SetupSignalHandler(context.Background())
+	metricsConfig.ServerTLSConfig = config.ServerTLSConfig
+	go metrics.Serve(ctx, metricsConfig)
+	config.MetricsRegisterer = metrics.Registry
 	_, err := endpoint.Listen(ctx, config)
 	if err != nil {
 		return err
--- a/pkg/drivers/generic/tx.go
+++ b/pkg/drivers/generic/tx.go
@@ -3,8 +3,11 @@ package generic
 import (
 	"context"
 	"database/sql"
+	"time"
 
+	"github.com/k3s-io/kine/pkg/metrics"
 	"github.com/k3s-io/kine/pkg/server"
+	"github.com/k3s-io/kine/pkg/util"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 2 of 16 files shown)


pkg/drivers/generic/metrics.go
```diff
@@ -0,0 +1,43 @@
...
+	SQLTime = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Name: "kine_sql_time_seconds",
+		Help: "Length of time per SQL operation",
+		Buckets: []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.15, 0.2, 0.25, 0.3, 0.35, 0.4, 0.45, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0,
+			1.5, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5, 5, 6, 7, 8, 9, 10, 15, 20, 25, 30},
+	})
+)
+
+const (
+	slowSQLThreshold = time.Second
```
brandond, line 30: This should probably be configurable somehow, rather than hardcoded. See https://github.com/k3s-io/kine/blob/dc877c3e2003aa6e01f1e35a09a6fb2578aa8a6e/pkg/drivers/jetstream/jetstream.go#L68

brandond (changes requested):
Looks good! Added a few comments about TLS and metrics selection.

I also notice that you didn't add any metrics support to the jetstream driver. Is this an intentional omission?

main.go
```diff
@@ -92,6 +101,8 @@ func run(c *cli.Context) error {
 		logrus.SetLevel(logrus.TraceLevel)
 	}
 	ctx := signals.SetupSignalHandler(context.Background())
+	go metrics.Serve(ctx, metricsAddr)
```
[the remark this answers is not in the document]
zqzten: Good catch, I missed it. I have made a seperate config for metrics server for future extensibility.

pkg/drivers/generic/metrics.go
```diff
@@ -0,0 +1,43 @@
...
+var (
+	SQLTotal = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "kine_sql_total",
+		Help: "Total number of SQL operations",
+	})
+
+	SQLErrors = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "kine_sql_errors_total",
+		Help: "Total number of SQL operation errors",
+	})
```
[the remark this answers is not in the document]
zqzten: Currently I cannot find a good way to reflect the SQL status code for now, maybe it's too low level and depends on specific drivers. I have added a `result` label to it (with value of `success` or `error` for now, maybe more in the future). I also improved the compaction metrics similarly.

zqzten:
@brandond Thanks for the comments! I have made changes accrodingly, PTAL.

For the JetStream driver, I omit it because it's not a SQL backend and cannot be included in the metrics introduced in this PR. Also, I'm not quite familiar with it so maybe we shall leave it's own metric implementations to some other one who really need them.

pkg/drivers/generic/metrics.go
```diff
@@ -0,0 +1,46 @@
+package generic
```
brandond, line 1: I'm not sure this file belongs in the generic driver. It doesn't seem to use anything from the generic package, and everything in this file is exported. It should probably be at `pkg/metrics/metrics.go`?

pkg/drivers/generic/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+const (
+	LabelSuccess = "success"
+	LabelError   = "error"
+)
+
+var (
+	SQLTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "kine_sql_total",
+		Help: "Total number of SQL operations",
+	}, []string{"result"})
```
brandond, line 19:
```suggestion
	}, []string{"error_code"})
```

pkg/drivers/generic/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+
+	SQLTime = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "kine_sql_time_seconds",
+		Help: "Length of time per SQL operation",
+		Buckets: []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.15, 0.2, 0.25, 0.3, 0.35, 0.4, 0.45, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0,
+			1.5, 2.0, 2.5, 3.0, 3.5, 4.0, 4.5, 5, 6, 7, 8, 9, 10, 15, 20, 25, 30},
+	}, []string{"result"})
```
brandond, line 26:
```suggestion
	}, []string{"error_code"})
```

pkg/drivers/generic/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func ObserveSQL(start time.Time, success bool, sql Stripped, args ...interface{}) {
+	result := LabelSuccess
+	if !success {
+		result = LabelError
+	}
+	SQLTotal.WithLabelValues(result).Inc()
+	duration := time.Since(start)
+	SQLTime.WithLabelValues(result).Observe(duration.Seconds())
```
brandond, line 42:
```suggestion
func ObserveSQL(start time.Time, code string, sql Stripped, args ...interface{}) {
	SQLTotal.WithLabelValues(code).Inc()
	duration := time.Since(start)
	SQLTime.WithLabelValues(code).Observe(duration.Seconds())
```

pkg/drivers/generic/tx.go
```diff
@@ -97,17 +98,29 @@ func (t *Tx) CurrentRevision(ctx context.Context) (int64, error) {
 	return id, err
 }
 
-func (t *Tx) query(ctx context.Context, sql string, args ...interface{}) (*sql.Rows, error) {
+func (t *Tx) query(ctx context.Context, sql string, args ...interface{}) (result *sql.Rows, err error) {
 	logrus.Tracef("TX QUERY %v : %s", args, Stripped(sql))
+	startTime := time.Now()
+	defer func() {
+		ObserveSQL(startTime, err == nil, Stripped(sql), args)
```
brandond, line 105:
```suggestion
		ObserveSQL(startTime, t.d.ErrCode(err), Stripped(sql), args)
```

brandond (changes requested):
Rather than just labeling these with success/failure, I think we can extract the actual error code/number? It'll need to be done on a per-driver basis, but if you add  `type ErrCode func(error) string` to generic.go, and `ErrCode ErrCode` to the Generic struct, you should be able to do something like this on each driver:

```golang
+       dialect.ErrCode = func(err error) string {
+               if err == nil {
+                       return ""
+               }
+               if err, ok := err.(*mysql.MySQLError); ok {
+                       return fmt.Sprint(err.Number)
+               }
+               return err.Error()
+       }
```

That's an example of mysql. postgres has err.Code, sqlite has err.ExtendedCode, and so on.

pkg/drivers/generic/metrics.go
zqzten: It use `Stripped` from the generic package to help print the slow SQL. Or maybe we can move `Stripped` to another general package like `utils`? wdyt

brandond: Ah right, yeah moving that outside the generic package seems like a good solution to allow reuse without import cycles.

zqzten:
@brandond Brilliant solution 👍 I have modifed the code accroding to your solution, PTAL.

zqzten:
@brandond Hi, are there any blocking issue on this PR? Seems that the test can't proceed with arm64 build for some unknown reason.

brandond:
Restarted CI

brandond (approved):

zqzten:
@brandond Mind merging this? We'd like to use this feature in community vendor, thanks!

PR #126, opened 2022-06-30
Only enforce RangeEnd + Limit restriction when RangeEnd is set

brandond:
Related to: https://github.com/rancher/rancher/issues/36922
Caused by: https://github.com/kubernetes/kubernetes/pull/108569

Get requests with a field-selector now create range requests that include a limit (page size) that is managed by the apiserver. Kine was enforcing an unconditional restriction on the limit not being set, when the apparent intention was to only prohibit use of limit + rangeEnd at the same time.

Kubernetes 1.24 will when searching within a namespace for a resource using FieldSelector on metatada.name, send a Get request for the fully qualified resource name, but with a non-zero limit:
`RangeRequest Key=/registry/services/specs/default/kubernetes RangeEnd= Limit=500 Revision=0 SortOrder=0 SortTarget=0 Serializable=false KeysOnly=false CountOnly=false`

Confirmed that this fixes use of fieldSelector with K3s 1.24.2

### DIFF
--- a/pkg/drivers/jetstream/jetstream.go
+++ b/pkg/drivers/jetstream/jetstream.go
@@ -280,7 +280,7 @@ func (j *JetStream) isKeyExpired(_ context.Context, createTime time.Time, value 
 }
 
 // Get returns the associated server.KeyValue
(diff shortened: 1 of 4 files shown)


briandowns (approved):

galal-hussein (approved):
Nice Catch! LGTM

PR #137, opened 2022-09-21
Bump nats client to v1.17.1-0.20220923

caleblloyd:
- Bumps the nats clinet to ~v1.17.0~ `v1.17.1-0.20220923204156-36d2b654c70f`
- Fixes the `DbSize` function to call Stream Info in order to get the stream size

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/lib/pq v1.10.2
 	github.com/mattn/go-sqlite3 v1.14.15
 	github.com/nats-io/jsm.go v0.0.31-0.20220317133147-fe318f464eee
-	github.com/nats-io/nats.go v1.13.1-0.20220318132711-e0e03e374228
+	github.com/nats-io/nats.go v1.17.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.11.0
 	github.com/rancher/wrangler v0.8.3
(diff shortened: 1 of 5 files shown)


brandond (approved):

brandond:
Hmm, it looks like the tests for the new nats server versions failed due to the DB test never succeeding: https://drone-pr.k3s.io/k3s-io/kine/94/1/3

Does the `DB_CONNECTION_TEST` need to be updated for the new versions?

brandond (changes requested):
Tests are failing

brandond:
No, it doesn't capture the actual output of the connection test. Best bet would probably be to start the database pod using the create command, and then test it with the test command.

We don't have an official support matrix; I kinda consider the minor versions that we have in the tests to be supported so if we're dropping v2.7 from the tests that says something.

caleblloyd:
I found the test issue, I had pasted an incorrect version into `DB_CONNECTION_TEST`; that is fixed

I am fine with keeping `2.7.4` in the tests...  I don't think we necessarily need to test every minor version of NATS, I don't want to bog down CI as NATS keeps publishing minor releases.  How does just testing `2.7.4` and the Current version of NATS sound?  Made that change too

brandond (approved):
https://drone-pr.k3s.io/k3s-io/kine/97/1/3

Ci failed but it looks like a flake; lgtm

Issue #87, opened 2021-05-25
Compaction not working on Cockroach DB?


afritzler:
I am running a `kube-apiserver` via `kine` on top of a CockroachDB. In the kine logs I see a lot of those errors:

```
time="2021-05-25T13:41:47Z" level=error msg="Compact failed: failed to compact to revision 82: pq: at or near \"where\": syntax error: unimplemented: this syntax"
time="2021-05-25T13:46:47Z" level=error msg="Compact failed: failed to compact to revision 290: pq: at or near \"where\": syntax error: unimplemented: this syntax"
time="2021-05-25T13:51:47Z" level=error msg="Compact failed: failed to compact to revision 498: pq: at or near \"where\": syntax error: unimplemented: this syntax"
time="2021-05-25T13:56:47Z" level=error msg="Compact failed: failed to compact to revision 705: pq: at or near \"where\": syntax error: unimplemented: this syntax"
```

I am running the latest master build of kine and CockroachDB version V21.1.1. `sql.defaults.serial_normalization` in my CockroachDB cluster is set to `sql_sequence`.

brandond:
Sounds like a new defect with cockroachdb?

afritzler:
Did somebody manage to run kine stable on cockroachDB? If so what version where you using?

brandond:
If you search for cockroachdb issues here and on the k3s repo, you should find a few discussions about it. We don't technically support it but a few folks have got it to work.

afritzler:
Ok it looks like this is the root cause of the compaction not working. https://github.com/cockroachdb/cockroach/issues/40963

afritzler:
So for me using the compaction query from the sqlite driver did the trick: https://github.com/afritzler/kine/commit/f3130cc9b3cf15a8dc69efc5b877177665012d96

brandond:
That's interesting, because on some releases of postgres the sqlite query won't execute at all...

PR #142, opened 2022-10-20
Add ability to enable remote profiling using pprof

radiohead:
### What

This PR adds a new flag, which, when enabled, allows for remote profiling using `net/http/pprof` package, with the handlers mounted to the metrics server.

By default the flag is disabled, to ensure that the change is backwards-compatible.

### Why

This allows for ad-hoc and continuous (e.g. using https://pyroscope.io/) profiling, which is useful for finding memory / cpu bottlenecks.

### DIFF
--- a/pkg/metrics/server.go
+++ b/pkg/metrics/server.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net"
 	"net/http"
+	"net/http/pprof"
 
 	"github.com/k3s-io/kine/pkg/tls"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -13,6 +14,7 @@ import (
 type Config struct {
 	ServerAddress   string
 	ServerTLSConfig tls.Config
+	EnableProfiling bool
 }
 
 const (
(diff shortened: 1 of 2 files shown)


radiohead:
Hey folks!

We're experimenting using kine with CockroachDB and I've run into OOMs with kine using up to 2GB of RAM in some situations. This seems to be related to https://github.com/k3s-io/kine/issues/87, since kine officially does not support CockroachDB.

That said, I'm looking into ways to improve peak memory usage, but I've noticed that there's no way to do remote profiling, so I've made this small change to enable it.

I've already tested this internally with https://hub.docker.com/repository/docker/radioheads/kine, but I'm happy to provide more info if necessary.

pkg/metrics/server.go
```diff
@@ -34,11 +36,21 @@ func Serve(ctx context.Context, config Config) {
 		logrus.Fatalf("error creating the metrics listener: %v", err)
 	}
 
+	mux := http.NewServeMux()
```
brandond, line 39: just out of curiosity, why did this get moved?

brandond (commented):
Looks good, thanks for the contribution! One small question.

pkg/metrics/server.go
radiohead: Whoops, this was not meant to be committed. Let me revert this particular change 🙂

radiohead: Reverted.

brandond (approved):

PR #143, opened 2022-10-21
Bump build/publish images and golang version

brandond:


.golangci.json
```diff
@@ -3,11 +3,28 @@
 		"disable-all": true,
 		"enable": [
 			"govet",
-			"golint",
-			"goimports"
+			"revive",
+			"goimports",
+      "gofmt"
```
briandowns, line 8: How ironic.

.golangci.json
```diff
@@ -3,11 +3,28 @@
...
+			{
+				"linters": "typecheck",
+				"text": "imported but not used"
+			},
+			{
+				"linters": "revive",
+				"text": "should have comment"
+			},
+			{
+				"linters": "revive",
```
briandowns, line 25: Why are we excluding this check?

.golangci.json
brandond: yay for per-file indentation conventions that my editor hides...

.golangci.json
brandond: I copied the exclude-rules from the k3s config. Let's see what happens if I don't exclude it.

briandowns (approved):

PR #146, opened 2022-11-18
Return a proper ResponseDeleteRange on Delete calls

### LINKED ISSUE #145: invalid DeleteRange response: "..." with 1.26.0-rc.0
### Summary

https://github.com/kubernetes/kubernetes/commit/421213b7a17b4d2753b3a00b7332a4bec3aba127 is now checking that the delete response range is valid. Kine seems to be returning an invalid response, which contains a `ResponseRange` instead of a `ResponseDeleteRange`, looking at https://github.com/k3s-io/kine/blob/master/pkg/server/delete.go#L30-L50

I've done some early testing with fixing the response format, but not sure if there is a blast radius that I am missing. Hoping to get some feedback from kine maintainers.

I'm creating a PR with my work, let's see where the discussion takes us.

neoaggelos:
Fixes #145

### DIFF
--- a/pkg/server/delete.go
+++ b/pkg/server/delete.go
@@ -37,10 +37,10 @@ func (l *LimitedServer) delete(ctx context.Context, key string, revision int64) 
 		Header: txnHeader(rev),
 		Responses: []*etcdserverpb.ResponseOp{
 			{
-				Response: &etcdserverpb.ResponseOp_ResponseRange{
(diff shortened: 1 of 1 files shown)


brandond (approved):
Thanks for the fix! This was likely just a case of copypasta that worked because nothing was actually checking the response.

PR #149, opened 2022-11-28
fix(mysql): quote database name in CREATE DATABASE statement

### LINKED ISSUE #148: mysql: creation of database with hyphen fails
Hey all,

when using kine with MySQL and a database name with a hyphen (e.g. `kine-test`), kine fails with

> time="2022-11-28T10:27:28Z" level=fatal msg="building kine: Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-test' at line 1"

because the database name is not quoted in the [CREATE DATABASE statement](https://github.com/k3s-io/kine/blob/v0.9.8/pkg/drivers/mysql/mysql.go#L146).

dakraus:
In order to allow the full Unicode Basic Multilingual Plane,
the database name is now quoted in the CREATE DATABASE statement.

fixes #148

### DIFF
--- a/pkg/drivers/mysql/mysql.go
+++ b/pkg/drivers/mysql/mysql.go
@@ -5,6 +5,7 @@ import (
 	cryptotls "crypto/tls"
 	"database/sql"
 	"fmt"
+	"strings"
 
 	"github.com/go-sql-driver/mysql"
 	"github.com/k3s-io/kine/pkg/drivers/generic"
@@ -137,7 +138,7 @@ func createDBIfNotExist(dataSourceName string) error {
 	if err != nil {
 		return err
 	}
-	dbName := config.DBName
+	dbName := quoteIdentifier(config.DBName)
 
 	db, err := sql.Open("mysql", dataSourceName)
 	if err != nil {
@@ -187,4 +188,8 @@ func prepareDSN(dataSourceName string, tlsConfig *cryptotls.Config) (string, err
 	parsedDSN := config.FormatDSN()
 
 	return parsedDSN, nil
+}
+
+func quoteIdentifier(id string) string {
+	return "`" + strings.ReplaceAll(id, "`", "``") + "`"
 }
--- a/pkg/drivers/pgsql/pgsql.go
+++ b/pkg/drivers/pgsql/pgsql.go
@@ -123,7 +123,7 @@ func createDBIfNotExist(dataSourceName string) error {
 		return err
 	}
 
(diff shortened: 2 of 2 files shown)


brandond:
Thank you for the contribution! Would you mind also fixing this for the other drivers as well?

dakraus:
Sure @brandond, I will check the other drivers to see if we this fix is needed elsewhere.

dakraus:
@brandond I added another commit to fix this for the PostgreSQL driver.  However this also means that the database names for the PostgreSQL driver are now case-sensitive according to the [official documentation](https://www.postgresql.org/docs/12/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS):

> Quoting an identifier also makes it case-sensitive, whereas unquoted names are always folded to lower case. 

Another thing I noticed while running the tests locally: The `CREATE DATABASE` statement is never executed for CockroachDB and kine fails during the schema creation with
```
[cockroachdb-v20.2] time="2022-11-29T12:51:42Z" level=info msg="starting metrics server path /metrics"
[cockroachdb-v20.2] time="2022-11-29T12:51:42Z" level=info msg="Configuring postgres database connection pooling: maxIdleConns=2, maxOpenConns=0, connMaxLifetime=0s"
[cockroachdb-v20.2] time="2022-11-29T12:51:42Z" level=info msg="Configuring database table schema and indexes, this may take a moment..."
[cockroachdb-v20.2] time="2022-11-29T12:51:42Z" level=trace msg="SETUP EXEC : CREATE TABLE IF NOT EXISTS kine ( id SERIAL PRIMARY KEY, name VARCHAR(630), created INTEGER, deleted INTEGER, create_revision INTEGER, prev_revision INTEGER, lease INTEGER, value bytea, old_value bytea );"
[cockroachdb-v20.2] time="2022-11-29T12:51:42Z" level=fatal msg="building kine: pq: no database or schema specified"
```

brandond:
Thanks! I take it the sqlite driver didn't need any changes?

> However this also means that the database names for the PostgreSQL driver are now case-sensitive.

That seems like a fair tradeoff; I'll make sure that is mentioned in the release notes.

> The CREATE DATABASE statement is never executed for CockroachDB

Hmm, the logs suggest that it is - are you saying that it actually is not? Or that the `CREATE TABLE` statement just silently fails?

dereknola:
Is this still relevant?

brandond:
It is, but I have been reluctant to merge it because of the change in behavior for Postgres.

PR #158, opened 2023-01-31
Add support for an embedded NATS server

bruth:


pkg/drivers/jetstream/jetstream.go
```diff
@@ -16,6 +18,7 @@ import (
 	"github.com/k3s-io/kine/pkg/server"
 	"github.com/k3s-io/kine/pkg/tls"
 	"github.com/nats-io/jsm.go/natscontext"
+	natsserver "github.com/nats-io/nats-server/v2/server"
```
brandond, line 21: What does this add to the total size of the binary?

Can we put this bit in a separate module that stubs out the embedded nats server so that we can opt into it with a build flag? K3s is very sensitive to stuff that increases the binary size.

bruth: Thanks @brandond. It increases the size by ~26% (7,768,648 bytes). Good suggestion to put this behind a build flag.

bruth:
@brandond Just an FYI that I will be a pushing a few more commits, refactoring the code, simplifying it, and adding tests where appropriate.

One question I did have was whether the CI that runs is the canonical/sufficient test suite that changes need to be tested against? Also one heads up is that I saw there was indeed a failure (which I am fixing) in the CI, but it still shows the CI passed: https://drone-pr.k3s.io/k3s-io/kine/147/1/3

bruth:
Sorry, one more question. There are various log statements being emitted within the NATS driver. Is there any policy/best practice recommended for logging given it will be embedded within k3s?

brandond:
> Just an FYI that I will be a pushing a few more commits, refactoring the code, simplifying it, and adding tests where appropriate.

Are you still working on this?

> whether the CI that runs is the canonical/sufficient test suite that changes need to be tested against? 

The CI is sufficient to confirm that it works with K3s, but our QA will still want to do a full set of tests on it when we update the Kine that is embedded in K3s itself.

> There are various log statements being emitted within the NATS driver. Is there any policy/best practice recommended for logging given it will be embedded within k3s?

Don't be overly chatty to the point where it obscures the K3s logs. Also keep in mind that K3s will never set the log level to TRACE, so if you ever want to see TRACE-level logs you'll need to run it standalone.


For the build flag, I would like to see the NATS server stuff stubbed out so that it just returns an error if someone attempts to use it when the embedded server is not enabled. I'm not sure we'll want to enable the embedded server in K3s, as 7MB is a non-trivial addition to the binary size.

bruth:
Just checking that I have picked up on this again and will have updates by tomorrow (Friday).

scripts/build
```diff
@@ -19,9 +19,9 @@ LINKFLAGS="-X github.com/k3s-io/kine/pkg/version.Version=$VERSION"
...
 echo Building Kine
-CGO_CFLAGS="-DSQLITE_ENABLE_DBSTAT_VTAB=1 -DSQLITE_USE_ALLOCA=1" go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o bin/kine
+CGO_CFLAGS="-DSQLITE_ENABLE_DBSTAT_VTAB=1 -DSQLITE_USE_ALLOCA=1" go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -tags nats -o bin/kine
 if [ "$CROSS" = "true" ] && [ "$ARCH" = "amd64" ]; then
-    GOOS=darwin go build -ldflags "$LINKFLAGS" -o bin/kine-darwin
-    GOOS=windows go build -ldflags "$LINKFLAGS" -o bin/kine-windows
-    GOOS=linux go build -ldflags "$LINKFLAGS" -o bin/kine-amd64
+    GOOS=darwin go build -ldflags "$LINKFLAGS" -tags nats -o bin/kine-darwin
+    GOOS=windows go build -ldflags "$LINKFLAGS" -tags nats -o bin/kine-windows
+    GOOS=linux go build -ldflags "$LINKFLAGS" -tags nats -o bin/kine-amd64
```
bruth, line 26: I presume this is only building the binaries for CI, correct? Otherwise if we want to have an opt-in embedded nats build using this tag, what is the precedent?

pkg/drivers/nats/server/interface.go
bruth, line 1: Put the NATS server dep in a separate package to differentiate with a build tag.

bruth (commented):
I paused my refactoring efforts and will put those into a subsequent PR. I will likely manifest as a rewrite of the backend to be more optimized.

The scope of this PR is now some config cleanup and additions, minor name changes for clarity (should be non-breaking, the existing kine URLs format/params are the same), and supporting an embedded server including a further opt-in for a pure socket between the client and server (no TCP). This would only be suitable for a single node deployment.

pkg/endpoint/endpoint.go
```diff
@@ -246,8 +246,8 @@ func getKineStorageBackend(ctx context.Context, driver, dsn string, cfg Config)
 		backend, err = pgsql.New(ctx, dsn, cfg.BackendTLSConfig, cfg.ConnectionPoolConfig, cfg.MetricsRegisterer)
 	case MySQLBackend:
 		backend, err = mysql.New(ctx, dsn, cfg.BackendTLSConfig, cfg.ConnectionPoolConfig, cfg.MetricsRegisterer)
-	case JetStreamBackend:
-		backend, err = jetstream.New(ctx, dsn, cfg.BackendTLSConfig)
+	case NATSBackend:
```
brandond, line 249: I think we probably need to continue to support the old `jetstream://` URI scheme, for compatibility purposes.
```suggestion
	case JetStreamBackend, NATSBackend:
```

pkg/endpoint/endpoint.go
```diff
@@ -27,13 +27,13 @@ import (
 )
 
 const (
-	KineSocket       = "unix://kine.sock"
-	SQLiteBackend    = "sqlite"
-	DQLiteBackend    = "dqlite"
-	ETCDBackend      = "etcd3"
-	JetStreamBackend = "jetstream"
```
brandond, line 34: Don't drop this; need to keep it for compat

scripts/build
brandond: This builds the artifacts that are used both for CI, and for release to GH and the Docker images. I am fine with having it on by default here; K3s embeds kine as a library and doesn't use the artifacts.

brandond (changes requested):

pkg/endpoint/endpoint.go
bruth: 👍 Good catch, I read the [comments in driver](https://github.com/k3s-io/kine/pull/158/files#diff-98f825ac9c2de7a75a0317fae315becccbf9c7dbff1a2d87444e15915dde4f81L58-L60) as it being `nats://` but you are right, it derives from the URL. Added back.

pkg/drivers/nats/server/server.go
```diff
@@ -0,0 +1,36 @@
...
+		// Parse the server config file as options.
+		var err error
+		opts, err = server.ProcessConfigFile(configFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to process NATS server config file: %w", err)
+		}
+	}
+
+	opts.DontListen = dontListen
+	opts.Port = 0
```
brandond, line 28: How do the DontListen and Port options interact? If you don't set the DontListen option, does it listen on a random port, or is there a default is uses instead?

bruth: That was an oversight, thanks. Port does not need to be reset if defined in the config. Internally if `DontListen` is set, `Port` is ignored. Fixed.

scripts/test-run-nats-embedded
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+start-test() {
+    local ip=$(cat $TEST_DIR/databases/*/metadata/ip)
+    local port=$(cat $TEST_DIR/databases/*/metadata/port)
+    KINE_IMAGE=$IMAGE KINE_ENDPOINT="nats://$ip:$port?embedServer" provision-kine
```
brandond, line 6: So just to be clear, the way to use this is to start kine with a `nats://` scheme datastore endpoint with the `embedServer` URI parameter set?

Can you also do something like `nats://127.0.0.1?embedServer`? If so, what port does it use?

If I set the `dontListen` option when using the embedded server, are the address and port still required, or can I just do `nats://?embedServer&dontListen`?

Can you add some examples that cover these different possibilities to a new markdown file in the [examples](https://github.com/k3s-io/kine/tree/master/examples) directory?

brandond (changes requested):

scripts/test-run-nats-embedded
bruth: Those are correct. There are default values, such as port 4222 and the `dontListen` example is also valid. I added some comments to the backend code, but will add examples.

gedw99: just adding myself as i want to try this out when the PR is merged..

bruth: Added an examples doc with more information.

gedw99: Thanks @bruth wil def help

examples/nats.md
```diff
@@ -0,0 +1,62 @@
...
+nats://nats-1.example.com?bucket=k3s,nats://nats-2.example.com,nats://nats-3.example.com
+```
+
+However, if a `contextFile` is feasible to specify, this is recommended rather than an inline comma-separated list.
+
+### Embedded
+
+- `nats://?embedServer`
+- `nats://?embedServer&serverConfig=/path/to/server.conf`
+- `nats://?embedServer&dontListen`
```
brandond, line 62: Thanks for adding docs!

Could you expand a bit on why someone might want to use the `dontListen` option? What's the default listen address and port? If it does listen, can other nodes connect to this node as the external NATS URL?

bruth: I can add to the doc, but will explain here first for context. We have had NATS users who need to embed the server and their environment is constrained to only exposing a controlled or limited set of ports. So they take advantage of using this non-TCP approach to establishing a connection between a local client and server in the same process.

brandond: sure that makes sense. I guess my question is, what is listening useful for? If you can't connect from another node when using the embedded server, why listen at all?

bruth: Right, this would only be useful in a single node setup embedded k3s -> kine -> nats.. which may be very niche, but an option.

The more common setup is embedded that does listen and can cluster and/or be configured for specialized setups (the two-node use case I helped work on during the SUSE Hackathon for instance). The specialized setup will require additional opt-in custom code for NATS, but will enable for active-passive or active-active setups.

scripts/test
```diff
@@ -26,7 +26,13 @@ echo "Did test-run-postgres $?"
...
-. ./scripts/test-run-jetstream
-echo "Did test-jetstream $?"
+. ./scripts/test-run-nats
+echo "did test-nats$?"
+
+. ./scripts/test-run-nats-embedded
+echo "did test-nats-embedded$?"
+
+. ./scripts/test-run-nats-socket
+echo "did test-nats-socket$?"
```
brandond, line 36: whitespace nit
```suggestion
echo "did test-nats $?"

. ./scripts/test-run-nats-embedded
echo "did test-nats-embedded $?"

. ./scripts/test-run-nats-socket
echo "did test-nats-socket $?"
```

scripts/test-run-nats-embedded
brandond, line 6: In the case of an embedded server there's no dedicated database container, right? So the IP and port shouldn't be set.
```suggestion
    KINE_IMAGE=$IMAGE KINE_ENDPOINT="nats://?embedServer" provision-kine
```

brandond (changes requested):

bruth:
I will rebase for the Go deps.

examples/nats.md
gedw99: Hey @bruth 

what is the embedded nats can be optionally put in leaf node and paired with a global nats cluster ? 

Then maybe devs can have their cake and eat it too:

either run in embedded standalone mode 

OR

run in embedded leaf mode paired up to a global nats cluster like Synadia etc .

is this a sensible option I wonder ?

bruth: Yes you could configure it that way by passing a config file that declares the `leafnodes` block pointing to leaf node *host*. Other than the single KV bucket used in the backend, none of the other NATS capabilities are used. However, if you had a reason to mirror that KV back up to the hub cluster or even another leaf, that could definitely be done.

I have not tried it, but in theory you could have a _read-only_ clone of another k3s+kine instance that uses a mirror of another instance. The second instance would receive all the KV changes over time (since its a stream under the hood) so the other k3s instance would mirror the same state (albeit lagging a tad since the replication is asynchronous).

examples/nats.md
```diff
@@ -0,0 +1,62 @@
...
+- `embedServer` - Specifies whether to embed a NATS server rather than relying on an external one. Default is `false`.
+- `serverConfig` - Specifies the path to a NATS server configuration file if `embedServer` is `true`. Defaults to nothing.
+- `dontListen` - Specifies whether the embedded NATS server should not rely on the TCP stack for a local client connection. Defaults to `false`
+- `stdoutLogging` - Specifies whether to log to STDOUT if `embedServer` is `true`. Default is false.
+
+## Examples
+
+### External
+
+- `nats://`
```
brandond, line 43: What do I get if I just do `nats://`? Does it expect to connect to an instance on localhost at the default port? Would it make more sense to have it default to a non-listening embedded server if no host is passed in the URI?

bruth: Happy to use this behavior if you are good with it. I think it makes sense and will be documented appropriately..

examples/nats.md
bruth: Assuming I understand correctly, if you wanted three kine instances with embedded nats, they could be clustered together via nats. If you wanted a kine instance/cluster hosting a mirror from another instance/cluster that is fine to.

The cluster is only necessary for the HA/FT, but you could have a single instance origin and single instance mirror if desired.

brandond:
I did a quick test of K3s with the nats server built in an it's actually not too terrible. 
```
// without nats
k3s binary dist/artifacts/k3s size 68452352 is less than max acceptable size of 73400320 bytes (70 MiB)
// with nats
k3s binary dist/artifacts/k3s size 69722112 is less than max acceptable size of 73400320 bytes (70 MiB)
```

I did notice that when

I did notice that even when running with `dontListen`, the embedded server still seems to be listening? thats what the log says at least. I am also confused by the fact that it reports that it's connecting to a server at `0.0.0.0` which isn't possible.

```console
brandond@dev01:~/go/src/github.com/k3s-io/k3s$ k3s server --token=token '--datastore-endpoint=nats://?embedServer&dontListen'
INFO[0000] Starting k3s v1.27.1+k3s-3988142f (3988142f)
INFO[0000] using an embedded NATS server
INFO[0000] started embedded NATS server
INFO[0000] using bucket: kine
INFO[0000] connecting to nats://0.0.0.0:4222
INFO[0000] Kine available at unix://kine.sock

```

examples/nats.md
bruth: Yes, not a problem.

bruth:
^Hrm.. will check out why it is still listening or reporting as such.

examples/nats.md
bruth: Addressed.

bruth:
Confirmed it was a misplaced log line, but does not actually listen.

examples/nats.md
bruth: Actually need to make a tweak.. since now specifying host:port is not enough to indicate an external server.

bruth:
@brandond If you are considering NATS can be built-in by default without requiring a build tag, then that may change the configuration design. Let me know either way and will update the code/docs to acknowledge that.

brandond:
I haven't run it by the rest of the team and our PM on the K3s side just yet. I think for Kine itself we can document it as available since we'll ship it that way in the release artifacts. Since the binary size impact for k3s is not *terrible* I think it should be doable.

bruth:
I updated the example doc and decided to use the `noEmbed` query param to disable use of an embedded server when the server in fact embedded. This will make the embedded build default to use an embedded server.. and a non-embedded build expect to connect to an external server.

brandond:
Ah, got it - so without the embedded server, the host:port are taken to be the address of an external server, and with it, they are the address to listen on?

bruth:
Yes, correct.

brandond (approved):
LGTM. Do you want to squash and re-push, or should I just do it when I merge?

bruth:
I appreciate the review and support!

PR #162, opened 2023-02-17
 switch to registry.k8s.io

DiptoChakrabarty:
The main Kubernetes container registry is now defaulted to registry.k8s.io. See: https://kubernetes.io/blog/2022/11/28/registry-k8s-io-faster-cheaper-ga/.
Switching to the new registry in the using with kubeadm sample.

### DIFF
--- a/examples/minimal.md
+++ b/examples/minimal.md
@@ -93,7 +93,7 @@ clusterName: kubernetes
 controllerManager: {}
 dns:
   type: CoreDNS
-imageRepository: k8s.gcr.io
+imageRepository: registry.k8s.io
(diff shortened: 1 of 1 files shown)


cwayne18:
Really appreciate this fix, could you please re-commit with -s to pass the dco check?

cwayne18:
oh you just did :D

cwayne18 (approved):

brandond:
Thanks! That kubeadm example config is essentially unmaintained, it could probably use some updating to use non-ancient versions of everything.

brandond (approved):

PR #161, opened 2023-02-16
Update readme + Restrict dependabot

dereknola:
- This updates the readme to include NATS Jetstream.
- This add a dependabot configuration file so dependabot STOPS TOUCHING go.mod and only looks at docker images... also it only runs weekly :smile:

PR #166, opened 2023-03-21
Lay groundwork for using updatecli

matttrach:
Add updatecli, have it update the go version on its own workflow, and add the /package dir to dependabot watchlist


matttrach:
This is the first iteration towards https://github.com/k3s-io/k3s/issues/7040.
If we want more on this I can continue on this branch, but this is useful and sets up the structure so I wanted to go ahead and share to get it on people's radar.

matttrach:
One benefit of this code is that it is useful to any repo we want to use updatecli with and is not specific to the repo, it could be added to any repo looking to use updatecli to setup the directory structure and show a simple example of it in action.

matttrach:
example of this in action: https://github.com/matttrach/kine/pull/1

updatecli/updatecli.d/updatecli.yml
```diff
@@ -0,0 +1,69 @@
...
+
+sources:
+  latestGoTag:
+    name: "Get latest Go 1.x tag"
+    kind: "gittag"
+    scmid: "go"
+    spec:
+      versionfilter:
+        kind: "regex"
+        pattern: 'go1\.[2-9][0-9]*\.\d$' # >= go1.20 < go2
```
matttrach, line 33: One quirk in the system is that unless your source uses proper semver then the sorting will be based on historical timeline rather than the tag numbers. The golang/go repo doesn't use proper semver (see the 'go' in front of the tag name), so it is not sorted alphabetically, but historically. The go1.19.7 tag was pushed after the go1.20.2 tag (the repo doesn't use GH releases), so the go1.19.7 tag was coming up as "latest", this regex makes sure that 1.20+ is selected, but we could run into the same issue when go1.30 is tagged.

updatecli/updatecli.d/updatecli.yml
```diff
@@ -0,0 +1,69 @@
...
+  kine:
+    kind: "github"
+    spec:
+      user: "{{ .github.user }}"
+      email: "{{ .github.email }}"
+      username: "{{ requiredEnv .github.username }}"
+      token: '{{ requiredEnv .github.token }}'
+      owner: "{{ .kine.org }}"
+      repository: "{{ .kine.repo }}"
+      branch: "{{ .kine.branch }}"
```
matttrach, line 13: This "branch" is a bit unintuitive, this is the branch that will be checked out from source and targeted with PRs, but a new randomly generated branch will be created to host the change. see https://github.com/matttrach/kine/pull/1 as an example

matttrach:
It seems like the Go linter is finding errors in Jetstream, this error doesn't seem to be related to my changes.

matttrach:
@macedogm I can't add you as a reviewer, but would like to know if you have any input on this.

macedogm:
@matttrach thanks for putting this PR. I started to review it and test the updates locally.

.github/workflows/updatecli.yml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.20.1'
```
macedogm, line 26: As I commented below in the UpdateCLI file, this will not work out of the box. To avoid such updates, we could use `stable` instead of the specific version. WDYT?

macedogm:
I guess we don't want to re-enable Dependabot updates for Go - https://github.com/k3s-io/kine/pull/161, right?

.github/workflows/updatecli.yml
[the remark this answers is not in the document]
macedogm: Note: this Go action is not needed for this repo in question, since UpdateCLI isn't using it to update Go debs.

updatecli/updatecli.d/updatecli.yml
matttrach: yes, I ran into this error as well and found that I needed to use the old un-scoped GH token style, which I assumed was being used on the runners. I don't have any insight into that environment, so thank you for the additional context! 
As for what to do about it... I guess having updatecli update its GH action file is not going to be viable, I will remove this file for now.

macedogm: Yes, unless K3s team has a bot account for such tasks.

macedogm: I would honestly like that GH could solve this issue. A bunch of people has the same problem and have to use one of those workarounds. 😮‍💨

.github/workflows/updatecli.yml
matttrach: yeah, I think setting this to stable would resolve the need to keep it up to date.

For future me, here is a link to the documentation about that: https://github.com/actions/setup-go#using-stableoldstable-aliases

.github/workflows/updatecli.yml
```diff
@@ -0,0 +1,36 @@
...
+
+permissions:
+  contents: write
+  issues: write
+  pull-requests: write
+
+jobs:
+  updatecli:
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main'
```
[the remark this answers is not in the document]
matttrach: yep, sorry. Thank you!

brandond (approved):

macedogm (approved):

Dockerfile.dapper
```diff
@@ -3,7 +3,7 @@ FROM golang:1.19-alpine3.16 AS dapper
 ARG ARCH=amd64
 
 RUN apk -U add bash coreutils git gcc musl-dev docker-cli vim less file curl wget ca-certificates
-RUN GOPROXY=direct go install golang.org/x/tools/cmd/goimports@gopls/v0.9.5
+RUN GOPROXY=direct go install golang.org/x/tools/cmd/goimports@gopls/latest
```
matttrach, line 6: I would like to highlight this change since it could introduce some instability (moving from a pinned version to latest for gopls)

matttrach: reverting this, CI failed

matttrach:
The new changes fix issues brought up in the CI, co-authored with @dereknola (really it was all him, I am just pushing it).
With this model in place, I will start to introduce opencli to other repos.

PR #168, opened 2023-03-22
merge into kine

AdamShannag:


brandond:
Thanks for the PR! We've always been a bit on the fence with regards to supporting non-open-source database engines, but we're willing to accept the PR even if we do not plan on supporting Oracle ourselves.

A couple housekeeping items:
* sign your commits for DCO
* rebase and squash so the only commits are for mod updates and addition of the new driver
* make sure that the new driver has test coverage and the tests pass
* confirm that we don't need any special licensing to use an Oracle database instance for testing

PR #175, opened 2023-04-17
Add config for golangci-lint

matttrach:
This adds a config to updatecli to keep golangci-lint up to date.

As I added this config I realized that having the repo info in a single file causes synchronization issues, so I removed the repo data from the values.yaml.
I want people to be able to work on different updatecli configs atomically and make changes without getting merge conflicts or having to synchronize pull requests, if you need to add your repo data to a single file that gets in the way of that goal. I think it is unlikely to cause any really painful issues to have the repo data in the config itself rather than the values.yaml.

### DIFF
--- a/updatecli/updatecli.d/dockerfile-dapper-golangci-lint.yml
+++ b/updatecli/updatecli.d/dockerfile-dapper-golangci-lint.yml
@@ -1,0 +1,64 @@
+# We don't want to actually pull the "latest" channel, we want to get the latest release and pin it
+#  when the PR comes in, we should be testing that this is actually what we want
+---
+name: "Update Dockerfile.dapper to pull the latest golangci-lint version for install"
+scms:
+  kine:
+    kind: "github"
+    spec:
+      user: "{{ .github.user }}"
+      email: "{{ .github.email }}"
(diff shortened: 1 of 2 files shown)


brandond (approved):

brandond:
@matttrach in reading the docs for the golangci-lint install script, it appears that not specifying a version argument to the install script always just installs the latest release from GitHub - so all we're versioning at the moment is the install script, which doesn't tend to change much or be tied to releases.

matttrach:
The install script's version matches the golangci-lint version, for instance when I use the v1.52.1 installer I am installing the golangci-lint version v1.52.1. 

I don't know what sources we trust for backwards compatibility and stability, so I assume that if there is a pinned version currently we probably want it pinned. Is golangci-lint a tool that we can trust to always have the latest release in the "latest" channel and for that release to be stable (not broken) and for it to be backwards compatible?

matttrach:
I hope these conditions are true, it is far simpler if we can trust latest and not worry about updating the version.

matttrach:
Based on a separate conversation, I am going to close this PR and move to using the "latest" version.

Issue #159, opened 2023-02-05
PostgreSQL's `target_session_attrs` is not supported


DerEnderKeks:
Currently, it is not possible to use a PostgreSQL cluster that is based on the native client failover mechanism using `target_session_attrs`, which is [implemented in libpq](https://www.postgresql.org/docs/15/libpq-connect.html#:~:text=Section%C2%A034.17.-,target_session_attrs,-This%20option%20determines), because Kine is using lib/pq, which doesn't implement this feature. 

Starting K3s with a PostgreSQL URL that contains the `target_session_attrs` parameter thus results in the following error:

```
level=fatal msg="starting kubernetes: preparing server: creating storage endpoint: building kine: pq: unrecognized configuration parameter \"target_session_attrs\""
```

As lib/pq is currently in maintenance mode, this feature probably won't happen there, and it would thus be required to replace lib/pq with a  driver that supports the `target_session_attrs` feature, preferably one that is based on libpq itself. Alternatively, something like [jackc/pgx](https://github.com/jackc/pgx) could be used, which is also written in pure Go, just like lib/pq, in case that this is a requirement.

Would it be feasible to do this? I would offer my help, if you are ok with replacing the driver with something else.

brandond:
If you're willing to take a shot at replacing it, please do so!

DerEnderKeks:
As the PR is merged, I'm closing this issue. As soon as the Kine version in K3s gets bumped, this problem should be solved there.

PR #179, opened 2023-04-19
Replace lib/pq with pgx

DerEnderKeks:
This PR replaces [lib/pq](https://github.com/lib/pq) with [pgx](https://github.com/jackc/pgx), resolving #159.

As pgx is compatible with the standard SQL interface, the change is rather simple and all tests passed.


brandond:
This looks good to me, thanks!

PR #192, opened 2023-06-19
Update NATS to v2.9.18

bruth:


brandond (approved):

brandond:
I see that CI has failed, is there anything interesting related to this change in the log?

bruth:
I forgot to update the actual test script for external NATS, so just did that.

I looked at the failure and tried to track down where/why it would occur, but impacts the older NATS server versions (2.7.x, and 2.8.x, no longer recommended to use). Is there a standard policy arounds backwards compat for kine drivers?

That said personally I want to understand what changed in, where I suspect, the Go NATS library causing the older server versions to fail..

PR #209, opened 2023-08-21
Fix SQLLog.compactor

hangscer8:
Because it will cause memory leak if we do not stop ticker when the function has completed.


brandond:
Sure, but this goroutine never exits unless the whole process is terminating. Have you run into any circumstances where this actual causes a leak of the timer?

hangscer8:
@brandond I think you are right, this pr just keeps code normal.

brandond (approved):

PR #218, opened 2023-10-13
enhance ttl events process logic

gshilei:
enhance ttl events process logic

### DIFF
--- a/pkg/logstructured/logstructured.go
+++ b/pkg/logstructured/logstructured.go
@@ -271,7 +271,7 @@ func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
 		rev, events, err := l.log.List(ctx, "/", "", 1000, 0, false)
 		for len(events) > 0 {
 			if err != nil {
-				logrus.Errorf("failed to read old events for ttl")
+				logrus.Errorf("failed to read old events for ttl: %v", err)
 				return
 			}
 
@@ -287,21 +287,14 @@ func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
 
 	go func() {
 		defer wg.Done()
-		for {
-			select {
-			case <-ctx.Done():
-				return
-			default:
-			}
-
-			for events := range l.log.Watch(ctx, "/") {
-				for _, event := range events {
-					if event.KV.Lease > 0 {
-						result <- event
-					}
+		for events := range l.log.Watch(ctx, "/") {
+			for _, event := range events {
+				if event.KV.Lease > 0 {
+					result <- event
 				}
 			}
 		}
+		logrus.Error("watch channel for ttl was closed due to slow processing")
 	}()
 
 	return result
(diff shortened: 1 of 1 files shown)


pkg/logstructured/logstructured.go
```diff
@@ -287,10 +287,18 @@ func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
 
 	go func() {
 		defer wg.Done()
-		for events := range l.log.Watch(ctx, "/") {
-			for _, event := range events {
-				if event.KV.Lease > 0 {
-					result <- event
+		for {
```
brandond, line 290: Can you describe the specific steps to reproduce the issue you're fixing with this change? 

I'm not understanding what adding an outer loop here that checks for context cancellation adds. The current `range l.log.Watch` should already break out of the loop when the context is cancelled, which will allow the goroutine to complete.

gshilei: Currently, the` l.log.Watch` is closed when ctx is cancelled and broadcaster distribute slowly. 

1. from the source code, we can see the watch invoke the `s.broadcaster.Subscribe(ctx, s.startWatch)`
image

2. each watch corresponds to a sub channel
image

3. broadcaster.stream will distribute the item, once the processing is slow, the channel also will be closed.
image

brandond: I will also note that the consumer of these events isn't great - I don't particularly like that it just starts up a sleeping goroutine to later wake and attempt to delete the key:
https://github.com/k3s-io/kine/blob/c5f2d892d70778fb9ad15208d4ca8617c3c794da/pkg/logstructured/logstructured.go#L305-L310

I have been meaning to improve this with a keyed work queue or something, but haven't got around to it.

brandond: The subscriber channel has space for 100 items, and the consumer operates in a goroutine within the same process; it should NOT be a slow consumer. Are you actually seeing it get dropped?

This would be pretty hard to catch at the moment, since it doesn't look like there are any logs to raise an error when the TTL deletion goroutines exit.

gshilei: I have also noticed that each goroutine to handle one event and need sleep and awake. For this issue, I have optimized it in my local code.  

But for this pr, it just to avoild the watch channel  being closed and ttl events will not be handled.

I add create_timestamp in the kine table. With this field, I solve the uncontrolled goroutine issue and add the compact strategy by time similar to etcd. The reason why I added this strategy is because kine default compact only keep latest 1000 item data. If cluster has large resources and include the delete item, the delete item may be comacted before it is be polled. 

If possible, I would like to submit the changes.

brandond: > kine default compact only keep latest 1000 item data 

That is not correct It keeps **AT LEAST 1000** revisions.

Kine observes the same strategy as the Kubernetes apiserver does when it compacts etcd. Every time a compact runs, it stores the current revision. The next time compaction runs, it compacts up to the previous revision. If the previous revision is less than 1000 revisions back (there have been less than 1000 changes since compaction was last run), it compacts 1000 revisions back instead.

This is all documented with references within the codebase; I'm not sure how you got the impression that it keeps only 1000 revisions.

https://github.com/k3s-io/kine/blob/c5f2d892d70778fb9ad15208d4ca8617c3c794da/pkg/logstructured/sqllog/sql.go#L102-L107

> If cluster has large resources and include the delete item, the delete item may be comacted before it is be polled.

Compaction is run every 5 minutes, so the poll would have to be delayed by at least 5 for the revision to be missed. Kubernetes clients should get a REVISION HAS BEEN COMPACTED error if this occurs, and restart their list-watch.

gshilei: Maybe my description is a bit incorrect，Kine keeps AT LEAST 1000 revisions.

The meaning I just wanted to express the details are follows:
Compaction is run every 5 minutes, it will check the compactRev and targetCompactRev. Ensure that we never compact the most recent 1000 revisions by invoke `safeCompactRev(targetCompactRev, currentRev)`. This is no problem.

If the cluster is under high pressure, the number of data inserted per second is much greater than 1000. Due to contractor and poll are two goroutines, the order cannot be controlled. Assuming 2000 new data are inserted and contains data with deleted equal to 1, what would happen if 'compact' occurs and then 'poll' ?

Above watch 'poll' will loss the date which delete equal to 1. Let's see the below comacted sql. the `kd.id < ?` which id may be include the data which are not be polled. 
image

brandond: > I add create_timestamp in the kine table

We generally try to avoid schema changes, as it is disruptive when upgrading multi-server clusters. I think just setting up a delaying queue from [workqueue](https://pkg.go.dev/k8s.io/client-go/util/workqueue), and pushing items into it to be processed later with `workqueue.AddAfter(key, delay)` would probably be the best approach.

gshilei: > We generally try to avoid schema changes, as it is disruptive when upgrading multi-server clusters. I think just setting up a delaying queue from [workqueue](https://pkg.go.dev/k8s.io/client-go/util/workqueue), and pushing items into it to be processed later with `workqueue.AddAfter(key, delay)` would probably be the best approach.

ok, i got it.

brandond (changes requested):
I think this is a good change, but the retry needs to be moved outside this goroutine.

If the watch channel is closed, we may miss events if all we do is re-watch. We need to instead both list (to find things that we missed) and watch (to find new events in the future).

This could probably be as simple as moving the retry loop into the `ttl(ctx context.Context)` function down below.

I would also like to see an error-level log message emitted when this occurs.

gshilei:
> I think this is a good change, but the retry needs to be moved outside this goroutine.
> 
> If the watch channel is closed, we may miss events if all we do is re-watch. We need to instead both list (to find things that we missed) and watch (to find new events in the future).
> 
> This could probably be as simple as moving the retry loop into the `ttl(ctx context.Context)` function down below.
> 
> I would also like to see an error-level log message emitted when this occurs.

I have moved the retry outside and add eventRecords map to filter the repeated events. 
Thank you for taking the time to review.

gshilei:
> I think this is a good change, but the retry needs to be moved outside this goroutine.
> 
> If the watch channel is closed, we may miss events if all we do is re-watch. We need to instead both list (to find things that we missed) and watch (to find new events in the future).
> 
> This could probably be as simple as moving the retry loop into the `ttl(ctx context.Context)` function down below.
> 
> I would also like to see an error-level log message emitted when this occurs.

Hi @brandond ,

Please help check whether the current changes are ok. If there are other problems, I will continue to make the changes you expect.  Thanks.

Issue #219, opened 2023-10-13
Watch channel of ttl events maybe abnormally closed 


gshilei:
Due to  subscribe channel maybe closed by slow processing，I enhance the ttl events logic to process watch events.

Otherwise, once the watch channel of ttl events is closed, ttl events will no longer be processed.

Thank you for taking the time to review.

https://github.com/k3s-io/kine/pull/218

gshilei:
#218

gshilei:
#218

Issue #47, opened 2020-07-29
Investigate removing goroutine-based TTL expiry model


brandond:
The current model of starting a goroutine for each entry with a TTL, and then waking the goroutine to attempt deletion, is non-optimal. The goroutines are not cancelled when a newer revision is create. It creates a large number of goroutines when k3s is starting up, including for revisions that are no longer current. It also resets the TTL on startup, since the creation time is not tracked. 

An alternative approach involving a dedicated TTL expiry goroutine may be more efficient. Fixing the TTL reset issue probably requires altering the schema to include a creation time.

PR #221, opened 2023-10-18
refactor: removing goroutine-based TTL expiry model

gshilei:
Hi @brandond,

According to your suggestion，I introduce the delaying work-queue to fix the "goroutine-based TTL expiry model" issue and the "watch channel of ttl events maybe abnormally closed" issue by the way.  Please help to review it.  if it make sense, then PR #218  can be closed in the future.

Related Issue:

- #47 
- #219 

In order to make the commit cleaner, please refer to new pr #235


pkg/logstructured/logstructured.go
```diff
@@ -294,28 +296,62 @@ func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
...
-				return
-			case <-time.After(time.Duration(event.KV.Lease) * time.Second):
-			}
-			mutex.Lock()
-			if _, _, _, err := l.Delete(ctx, event.KV.Key, event.KV.ModRevision); err != nil {
-				logrus.Errorf("failed to delete expired key: %v", err)
-			}
-			mutex.Unlock()
-		}(event)
+	queue := workqueue.NewRateLimitingQueue(workqueue.DefaultItemBasedRateLimiter())
```
brandond, line 310: I don't think we want to put a rate limit on things expiring, do we? Would `NewDelayingQueue()` be more appropriate?

pkg/logstructured/logstructured.go
```diff
@@ -294,28 +296,62 @@ func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
...
+	go func() {
+		for l.handleTTLEvents(ctx, queue, 3) {
```
brandond, line 312: Was 3 an arbitrary choice? Is there any reason we would want to skip deleting an expired item after a certain number of attempts?

brandond (changes requested):

pkg/logstructured/logstructured.go
gshilei: > Was 3 an arbitrary choice? Is there any reason we would want to skip deleting an expired item after a certain number of attempts?

Since the error information was not processed in the previous logic, three retries are temporarily added here. Maybe, we can not limit the number of times. If it fails, just put it back into the queue. What's your suggestion?

https://github.com/k3s-io/kine/blob/c5f2d892d70778fb9ad15208d4ca8617c3c794da/pkg/logstructured/logstructured.go#L302C1-L318C3

pkg/logstructured/logstructured.go
gshilei: NewRateLimitingQueue support delaying queue with method queue.AddAfter and offer method queue.AddRateLimited

brandond: My question was why we want to rate limit at all. Things should be deleted as they expire. I do not understand why we would want to rate limit the timely removal of deleted items. Can you explain why that would be desired?

pkg/logstructured/logstructured.go
brandond: If the error is that the revision is too old, it can be removed from the queue as the key has been updated. If there is some other error with the database or network, it should probably be requeued?

pkg/logstructured/logstructured.go
gshilei: Due to RateLimitingQueue support all functions of delaying queue.  event is put into queue with fixed event lease interval. **When event error deleting, here just use it rate limit function, the rate limit function provide exponential retries**. For the error retrying, Maybe, we also can use delayingQueue with fixed interval.

`// NewRateLimitingQueue constructs a new workqueue with rateLimited queuing ability
// Remember to call Forget!  If you don't, you may end up tracking failures forever.
// NewRateLimitingQueue does not emit metrics. For use with a MetricsProvider, please use
// NewNamedRateLimitingQueue instead.
func NewRateLimitingQueue(rateLimiter RateLimiter) RateLimitingInterface {
	return &rateLimitingType{
		DelayingInterface: NewDelayingQueue(),
		rateLimiter:       rateLimiter,
	}
}
`

brandond: You're not making a case for using exponential retries or rate limiting of queue entries. Why do you want that?  I don't think those are things we want. I think a simple delaying queue is more appropriate. Can you explain why that is not the case?

pkg/logstructured/logstructured.go
gshilei: I agree with your point of view, then I will remove the number limits.

pkg/logstructured/logstructured.go
gshilei: I got it. I will replace the RateLimitingQueue with DelayingQueue later.

gshilei: > You're not making a case for using exponential retries or rate limiting of queue entry processing. Why do you want that? I don't think those are things we want. I think a simple delaying queue is more appropriate. Can you explain why that is not the case?

Fixed is done.

pkg/logstructured/logstructured.go
```diff
@@ -294,28 +300,57 @@ func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
...
+		for event := range l.ttlEvents(ctx) {
+			queue.AddAfter(event, time.Duration(event.KV.Lease)*time.Second)
+		}
+	}
+}
+
+func (l *LogStructured) handleTTLEvents(ctx context.Context, queue workqueue.DelayingInterface) bool {
+	item, shutdown := queue.Get()
+	if shutdown {
+		logrus.Info("ttl events work queue has shut down")
```
brandond, line 337: Log messages should generally be capitalized like a sentence. Error messages should not.
```suggestion
		logrus.Info("TTL events work queue has shut down")
```

pkg/logstructured/logstructured.go
```diff
@@ -294,28 +300,57 @@ func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
...
+		return false
+	}
+
+	defer queue.Done(item)
+	event, ok := item.(*server.Event)
+	if !ok {
+		return true
+	}
+
+	logrus.Tracef("handle ttl events key: %v, modRev: %v", event.KV.Key, event.KV.ModRevision)
```
brandond, line 347:
```suggestion
	logrus.Tracef("TTL events key: %v, modRev: %v", event.KV.Key, event.KV.ModRevision)
```

pkg/logstructured/logstructured.go
```diff
@@ -294,28 +300,57 @@ func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
...
+
+	if _, _, _, err := l.Delete(ctx, event.KV.Key, event.KV.ModRevision); err != nil {
```
brandond, line 349: You should manually test to see what happens if the key is deleted or updated before the TTL expires, I think the current state of retrying on all errors is incorrect. I think you need to handle `ErrKeyExists` and maybe also `ErrCompacted` here by not retrying. It is possible that the key was already deleted, or it was updated to no longer have a TTL.

pkg/logstructured/logstructured.go
```diff
@@ -5,8 +5,14 @@ import (
...
 
-	"github.com/k3s-io/kine/pkg/server"
 	"github.com/sirupsen/logrus"
+	"k8s.io/client-go/util/workqueue"
+
+	"github.com/k3s-io/kine/pkg/server"
+)
+
+const (
+	retryInterval = 5 * time.Second
```
brandond, line 15: This could probably be shorter - maybe start out at 250ms? Just kind of guessing.

brandond (changes requested):

pkg/logstructured/logstructured.go
gshilei: Yeah, I have already tested for different cases. Also, let's explore the logic of of Delete from a code perspective，I manually added comments in the following code.

```
func (l *LogStructured) Delete(ctx context.Context, key string, revision int64) (revRet int64, kvRet *server.KeyValue, deletedRet bool, errRet error) {
	defer func() {
		l.adjustRevision(ctx, &revRet)
		logrus.Tracef("DELETE %s, rev=%d => rev=%d, kv=%v, deleted=%v, err=%v", key, revision, revRet, kvRet != nil, deletedRet, errRet)
	}()
        
        // Case 1: Retry
        // Obtain the specified key from the backend, and if it fails, return error. 
        // This error may contain network or database errors.
	rev, event, err := l.get(ctx, key, "", 1, 0, true)
	if err != nil {
		return 0, nil, false, err
	}
         
        // Case 2: Don't Retry
        // If the key don't exist in DB,  event will be nil. So the returned error is nil.
	if event == nil {
		return rev, nil, true, nil
	}
        
        // Case 3: Don't Retry
        // If the event is already deleted, the deleted column in DB will be set to 1.
        // When trying to delete again, the returned error is nil. 
	if event.Delete {
		return rev, event.KV, true, nil
	}
        
        // Case 4: Don't Retry
        // If the event is updated, the updated event modRevision will be not equal to the previous revision.
        // So the returned error is nil.
	if revision != 0 && event.KV.ModRevision != revision {
		return rev, event.KV, false, nil
	}

	deleteEvent := &server.Event{
		Delete: true,
		KV:     event.KV,
		PrevKV: event.KV,
	}

	rev, err = l.log.Append(ctx, deleteEvent)
	if err != nil {
		// If error on Append we assume it's a UNIQUE constraint error, so we fetch the latest (if we can)
		// and return that the delete failed
                 
                // Case 5: Don't Retry (Here, I have some questions.)
		latestRev, latestEvent, latestErr := l.get(ctx, key, "", 1, 0, true)
                 // Q1: If latestErr != nil or latestEvent == nil, Why return deleted == false and error == nil ?
                 // For this case:
                 //       If latestErr != nil, do we return latestErr ?
                 //       if llatestErr == nil and latestEvent == nil, do we already return error with err?
		if latestErr != nil || latestEvent == nil {
			return rev, event.KV, false, nil
		}
                // Q2:  If the event was already deleted, why also return deleted == false and error == nil ?
                // For this case:
                //         do we need return deleted == true and error == nil ?
		return latestRev, latestEvent.KV, false, nil
	}
        // Case 6:  Don't Retry.
        // Maybe it will only be returned when error == nil
	return rev, event.KV, true, err
}

```

pkg/logstructured/logstructured.go
```diff
@@ -294,28 +300,57 @@ func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
 				}
 			}
 		}
+
+		if ctx.Err() == nil {
+			logrus.Error("ttl events watch channel was closed due to slow processing")
```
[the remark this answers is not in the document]
gshilei: Fixed.

pkg/logstructured/logstructured.go
gshilei: Fixed

pkg/logstructured/logstructured.go
gshilei: Fixed

pkg/logstructured/logstructured.go
brandond: right, it's been a while since I did a deep dive into that bit of the code. You are correct that it appears that the only thing that should return `errRet != nil` would be an error on the initial get. All other paths swallow the error and just return `deletedRet = false` to indicate that the key was not deleted. So I guess we just want to requeue on any error.

pkg/logstructured/logstructured.go
```diff
@@ -294,28 +300,55 @@ func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
...
+	defer queue.Done(item)
+	event, ok := item.(*server.Event)
+	if !ok {
+		return true
+	}
+
+	logrus.Tracef("TTL events key: %v, modRev: %v", event.KV.Key, event.KV.ModRevision)
+
+	if _, _, _, err := l.Delete(ctx, event.KV.Key, event.KV.ModRevision); err != nil {
+		logrus.Errorf("failed to delete expired key: %v", err)
```
brandond, line 348:
```suggestion
		logrus.Errorf("Failed to delete expired key: %v", err)
```

brandond (changes requested):
Final nit. lgtm otherwise. thanks for talking through it with me!

pkg/logstructured/logstructured.go
gshilei: Fixed.

gshilei:
> Final nit. lgtm otherwise. thanks for talking through it with me!

Thank you very much for your more professional attitude towards this review, Hope you have a good day.

pkg/logstructured/logstructured.go
```diff
@@ -294,28 +300,55 @@ func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
...
+	for {
+		select {
+		case <-ctx.Done():
+			queue.ShutDown()
+			return
+		default:
+		}
+
+		for event := range l.ttlEvents(ctx) {
+			queue.AddAfter(event, time.Duration(event.KV.Lease)*time.Second)
```
brandond, line 327: I was thinking about this a bit more, and realized that we are missing a bit of nuance about how workqueues are normally used alongside an object cache.

We don't actually want to enqueue the full event here, just the event key. If we enqueue the event, then each update is queued for processing. We want to enqueue the key, so that each update to the TTL pushes out its time in the queue. By doing this, we can mostly avoid deleting old revisions entirely.

This does mean that we need to maintain a store parallel to the queue, that hold the most recently seen revision for each key. When the key does come up in the queue, we delete `key, revision` from the datastore, and remove the key from both the queue and the store.

[cache.NewStore()](https://pkg.go.dev/k8s.io/client-go/tools/cache#NewStore) is probably a good choice for this very simple use case.

gshilei: I understand what you mean. Your idea is actually to ignore intermediate versions of updated events, which would be more efficient. I think maybe we don't need to introduce cache.ThreadSafeStore directly in this simple case. Maybe go's native sync.Map can also solve this problem?

Firstly, we can't just enqueue the key into the queue, we need to put the struct{key, revision} into the queue. Because when an event is updated, we cannot directly update the enqueue time of the previous revision in the workqueue to the latest version. Therefore, after getting the event from the queue, we need to compare whether the key of the event exists in the map. If it exists, compare whether the revision is smaller than the latest revision in the map. If so, then we can ignore the event and avoid doing the delete operation.

What's your suggestion?

brandond: > I think maybe we don't need to introduce cache.ThreadSafeStore directly in this simple case. Maybe go's native sync.Map can also solve this problem

I'm not sure, I'd have to take a closer look at it. I know that Kubernetes uses the cache library alongside workqueue, and I would expect there to be some specific reason for that.

> we cannot directly update the enqueue time of the previous version in the workqueue to the latest version as delaying queue only choose the minimum time.

Hmm, that is unfortunate. I was expecting workqueue to allow us to delay processing of an item by reenqueuing it with a higher delay. If this is not the case then some of its advantages are lost.

I am also realizing that the fact that we simultaneously list and watch will create problems. We need to either list then start a watch at the last revision seen by the watch (which is what Kubernetes does), or use a store type which understands key revisions. There is such a store available in the cache module, but it may be more correct to just list-then-watch.

gshilei: >> I think maybe we don't need to introduce cache.ThreadSafeStore directly in this simple case. Maybe go's native sync.Map can also solve this problem

> I'm not sure, I'd have to take a closer look at it. I know that Kubernetes uses the cache module alongside workqueue, and I would expect there to be some specific reason for that.

image

As simply shown in the figure above, workqueue and ThreadSafeStore are mainly used together for apiserver List-Watch processing logic in client-go. The main function of ThreadSafeStore is to store objects from list-watch and maintain complex indexes of objects and keys. Through the encapsulation of client-go, the desired object can be obtained indirectly from ThreadSafeStore according to the specified key or rule through Get/List and other interfaces in the custom controller.
In our scenario, we do not need to maintain the mapping from Object to key, and encapsulate the Get/List interface to query Object. In addition, WorkQueue and ThreadSafeStore are inherently independent, and there is no binding relationship when used. We only store the latest Revision corresponding to the Event's Key.

>> we cannot directly update the enqueue time of the previous version in the workqueue to the latest version as delaying queue only choose the minimum time.
> Hmm, that is unfortunate. I was expecting workqueue to allow us to delay processing of an item by reenqueuing it with a higher delay. If this is not the case then some of its advantages are lost.

The design of client-go's delaying work queue itself is to process the Object received from apiserver List-Watch. This design takes into account the final reconciliation consistency of the Object. When there is already an Item in the queue, the same Item will be re-enqueued. , he will compare the delay times of the two Items and choose the smallest time in order to process the Item faster. Reference code:
https://github.com/kubernetes/client-go/blob/33d14001dbdfd2f3eccfa907b8cc3738f514680e/util/workqueue/delaying_queue.go#L315C3-L318C4

> I am also realizing that the fact that we simultaneously list and watch will create problems due to old and new revisions being placed into the channel simultaneously. We need to either list then start a watch at the last revision seen by the watch (which is what Kubernetes does), or use a store type which understands key revisions. There is such a store available in the cache module, but it may be more correct to just list-then-watch.

I also noticed the problem of List and watch in the ttlEvents function of the current code. Although the best way is to adjust the order of List and Watch to ensure the order of elements placed in the delaying work queue. This problem only occurs when Kine is restarted. As a compromise method, a parallel store is introduced, the type is sync.Map, which stores the latest key and revision mapping from List and Watch. When the delaying work queue dequeues an element which contains {key, revision}, we need to compare whether the revision of the element is equal to the revision corresponding to the element in the store. If it is equal, then we actually call the Delete method to delete the element.

https://github.com/k3s-io/kine/blob/c5f2d892d70778fb9ad15208d4ca8617c3c794da/pkg/logstructured/logstructured.go#L269C1-L297C5

gshilei: Hi @brandond，
what do you think for above discussion ?

brandond: >  When there is already an Item in the queue and the same Item be re-enqueued, it will compare the delay times of the two Items and choose the smallest time in order to process the Item faster.

It's unfortunate that the comparison here isn't a configurable property of the queue. It's probably not worth maintaining our own queue type just to change this:
https://github.com/kubernetes/client-go/blob/33d14001dbdfd2f3eccfa907b8cc3738f514680e/util/workqueue/delaying_queue.go#L312

If we store key+revision, then we are missing a lot of the benefits of the queue, as each key+revision will be a unique entry and still need to be processed. There will be no such thing as an item "already being in the queue" as every key+revision is guaranteed to be unique.

I think instead we should enqueue only the key, and maintain a map something like this:
```golang
type ttlKV struct {
  key string
  modRevision int64
  expires time.Time
}

keyMap := map[string]ttlKV
```

When we get an event, the TTL duration is added to the current time, to calculate `expires` for the map entry. When the key is dequeued, check if the expiration time has been reached. If so, delete the key+revision; if not re-enqueue the key.

> This problem only occurs when Kine is restarted.

I believe this is a problem any time the watch channel is closed and `ttlEvents` exits the loop, not just when kine is restarted. When this happens, we want to run `ttlEvents` again which will perform a new the list+watch, to ensure that we didn't miss anything. This issue is explicitly one of the things that you were hoping to fix, correct?

gshilei: Yes, I agree with your point of view. I will fix these problems later.

gshilei: Hi @brandond, 

I have updated, please help to have a review. Thanks.

gshilei: Hi @brandond,

I have updated this PR, please take a time to review. thanks.

pkg/logstructured/logstructured.go
```diff
@@ -262,8 +268,85 @@ func (l *LogStructured) Update(ctx context.Context, key string, value []byte, re
...
+
+	eventKV := loadTTLEventKV(ttlEventMap, key)
+	if eventKV == nil {
+		logrus.Errorf("Failed to find ttl event for key %v", key)
+		return true
+	}
+
+	if eventKV.expiredAt.After(time.Now()) {
+		logrus.Tracef("TTL event key %v has not expired yet, the latest expiration time is %v, requeuing", key, eventKV.expiredAt)
+		queue.AddAfter(key, eventKV.expiredAt.Sub(time.Now()))
```
brandond, line 326: nit: use the idiomatic `Until`
```suggestion
		queue.AddAfter(key, time.Until(eventKV.expiredAt))
```

pkg/logstructured/logstructured.go
```diff
@@ -262,8 +268,85 @@ func (l *LogStructured) Update(ctx context.Context, key string, value []byte, re
...
+		return true
+	}
+
+	logrus.Tracef("Delete ttl event key %v, modRev %v", eventKV.key, eventKV.modRevision)
+	_, _, _, err := l.Delete(ctx, eventKV.key, eventKV.modRevision)
+	curEventKV := loadTTLEventKV(ttlEventMap, key)
+	if curEventKV.expiredAt.After(eventKV.expiredAt) {
+		logrus.Infof("TTL event key %v has updated, requeue", key)
+		queue.AddAfter(key, curEventKV.expiredAt.Sub(time.Now()))
```
brandond: nit: use the idiomatic `Until`
```suggestion
		queue.AddAfter(key, time.Until(curEventKV.expiredAt))
```

pkg/logstructured/logstructured.go
```diff
@@ -262,8 +268,85 @@ func (l *LogStructured) Update(ctx context.Context, key string, value []byte, re
...
+			expiredAt := time.Now().Add(time.Duration(event.KV.Lease) * time.Second)
+			eventKV := loadTTLEventKV(ttlEventMap, event.KV.Key)
+			if eventKV == nil {
+				storeTTLEventKV(ttlEventMap, event.KV.Key, event.KV.ModRevision, expiredAt)
+				queue.AddAfter(event.KV.Key, time.Duration(event.KV.Lease)*time.Second)
+			} else {
+				if event.KV.ModRevision > eventKV.modRevision {
+					storeTTLEventKV(ttlEventMap, event.KV.Key, event.KV.ModRevision, expiredAt)
+				}
+			}
```
brandond: Question: is there any possibility of a race condition here? What happens if the queue worker removes a kv from the map in the time between it is returned from `loadTTLEventKV`, and when we check to see if it is older than the current kv? I think that should be fine?

Nits:
* simplify this by having `storeTTLEventKV` take the whole KV, instead of getting the key, revision, and ttl for it.
* make `storeTTLEventKV` return the TTL duration so that we can enqueue using it.
* enqueue the key even if there is an existing event in the queue, it's possible the new event has a lower TTL and needs to fire **before** the existing queued event (for example, 60 minute ttl updated to 5 seconds; existing queued item would fire too late).
```suggestion
			existingKV := loadTTLEventKV(ttlEventMap, event.KV.Key)
			if existingKV == nil {
				expires := storeTTLEventKV(ttlEventMap, event.KV)
				queue.AddAfter(event.KV.Key, expires)
			} else if event.KV.ModRevision > existingKV.modRevision {
				expires := storeTTLEventKV(ttlEventMap, event.KV)
				queue.AddAfter(event.KV.Key, expires)
			}
```

pkg/logstructured/logstructured.go
```diff
@@ -262,8 +268,85 @@ func (l *LogStructured) Update(ctx context.Context, key string, value []byte, re
...
+	curEventKV := loadTTLEventKV(ttlEventMap, key)
+	if curEventKV.expiredAt.After(eventKV.expiredAt) {
+		logrus.Infof("TTL event key %v has updated, requeue", key)
+		queue.AddAfter(key, curEventKV.expiredAt.Sub(time.Now()))
+		return true
+	}
+	if err != nil {
+		logrus.Errorf("Failed to delete expired event key %v, %v, requeuing", eventKV.key, err)
+		curEventKV.expiredAt = time.Now().Add(retryInterval)
+		ttlEventMap.Store(key, curEventKV)
```
brandond: You don't need to update the TTL on the stored event, just leave it in the past and re-enqueue after the retry interval.
```suggestion
		logrus.Errorf("Failed to delete key %v at end of lease: %v, requeuing", eventKV.key, err)
```

pkg/logstructured/logstructured.go
```diff
@@ -274,81 +357,65 @@ func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
...
-	logrus.Tracef("TTL events key: %v, modRev: %v", event.KV.Key, event.KV.ModRevision)
+	return eventKV
+}
 
-	if _, _, _, err := l.Delete(ctx, event.KV.Key, event.KV.ModRevision); err != nil {
-		logrus.Errorf("Failed to delete expired key: %v, requeueing", err)
-		queue.AddAfter(item, retryInterval)
-	}
-	return true
+func storeTTLEventKV(ttlEventMap *sync.Map, key string, modRevision int64, expiredAt time.Time) {
```
brandond: make `storeTTLEventKV` take the whole KV, instead of getting the key, revision, and ttl for it.
make` storeTTLEventKV` return the TTL duration so that we can enqueue using it.

pkg/logstructured/logstructured.go
```diff
@@ -274,81 +357,65 @@ func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
 
 	go func() {
 		defer wg.Done()
```
brandond, line 370: Make sure to close the channel. If you don't do this and there is an error in the for loop that returns early, the read from the channel will block forever.
```suggestion
		defer close(lastListRevision)
		defer wg.Done()
```

pkg/logstructured/logstructured.go
```diff
@@ -274,81 +357,65 @@ func (l *LogStructured) ttlEvents(ctx context.Context) chan *server.Event {
...
 
 			_, events, err = l.log.List(ctx, "/", events[len(events)-1].KV.Key, 1000, rev, false)
 		}
+		lastListRevision <- lastRev
 	}()
 
 	go func() {
 		defer wg.Done()
-		for events := range l.log.Watch(ctx, "/") {
+		revision := <-lastListRevision
```
brandond, line 396: check if the channel read returned 0 - if so then the list failed and we should error out here too
```suggestion
		revision := <-lastListRevision
```

brandond (changes requested):

pkg/logstructured/logstructured.go
gshilei: >  Question: is there any possibility of a race condition here? What happens if the queue worker removes a kv from the map in the time between it is returned from loadTTLEventKV, and when we check to see if it is older than the current kv? I think that should be fine?

I have already noticed it. If use sync.map, there is a race. So I use mutex + map to avoid.

brandond (changes requested):

pkg/logstructured/logstructured.go
```diff
@@ -291,16 +291,16 @@ func (l *LogStructured) ttl(ctx context.Context) {
 				continue
 			}
 
-			expiredAt := time.Now().Add(time.Duration(event.KV.Lease) * time.Second)
 			eventKV := loadTTLEventKV(rwMutex, ttlEventKVMap, event.KV.Key)
 			if eventKV == nil {
 				logrus.Tracef("Add ttl event key %v, modRev %v", event.KV.Key, event.KV.ModRevision)
-				storeTTLEventKV(rwMutex, ttlEventKVMap, event.KV.Key, event.KV.ModRevision, expiredAt)
-				queue.AddAfter(event.KV.Key, time.Duration(event.KV.Lease)*time.Second)
+				expires := storeTTLEventKV(rwMutex, ttlEventKVMap, event.KV)
```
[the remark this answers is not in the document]
gshilei: > just passing it in here during the store won't fix anything; you'd need to get the lock before loading the the entry, and hold it until after the checks are done.
> 
> I'm not even convinced that it's necessary, I just wanted to think about whether or not this creates the possibility of a race condition.
> 
> If it does matter, and we need to hold a lock while checking the existing entry, then we probably need to make a type that holds both the map and the lock, and have a function on that type to handle locking and comparison - rather than exposing it here.

I don't think there will be any problems here. First of all, the event processing logic is handled by a single process, Secondly, even if the load retrieves the previous deleted element, the event. ModRevision will be compared during the update. If it is a new event, the new event will be queued.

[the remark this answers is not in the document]
brandond: I agree, I think we're probably fine without a lock.

gshilei:
Hi @brandond,

Do you have any other suggestions for this PR? In addition, I found that "continuous-integration/drone/pr build" was killed  abnormally. How can I trigger it again?

pkg/logstructured/logstructured.go
```diff
@@ -256,66 +268,164 @@ func (l *LogStructured) Update(ctx context.Context, key string, value []byte, re
 	return rev, updateEvent.KV, true, err
 }
 
+func (l *LogStructured) ttl(ctx context.Context) {
+	// vary naive TTL support
```
brandond, line 272: you can drop this comment now, I think ;)

brandond (changes requested):
One nit on a comment that can be removed.

Can you rebase and squash this? Once that's set I think we're good.

PR #241, opened 2023-11-03
Bugfix? create pgql database if not exists

oz123:
With the current master branch code
(commit afb2121bb0d8c2e079f90f7d3a1b1a2c3338bfa1)
I see the following error:

root@vm2:/home/vagrant# kine --debug --endpoint postgres://postgres:postgresql@localhost:5432/foobster INFO[2023-11-03T20:07:09.379099258Z] metrics server is starting to listen at :8080 INFO[2023-11-03T20:07:09.379709686Z] starting metrics server path /metrics FATA[2023-11-03T20:07:09.40247106Z] building kine: failed to connect to `host=localhost user=postgres database=foobster`: server error (FATAL: database "foobster" does not exist (SQLSTATE 3D000))

This is fixed with the current commit.

### DIFF
--- a/pkg/drivers/pgsql/pgsql.go
+++ b/pkg/drivers/pgsql/pgsql.go
@@ -134,34 +134,28 @@ func createDBIfNotExist(dataSourceName string) error {
 	}
 
 	dbName := strings.SplitN(u.Path, "/", 2)[1]
-	db, err := sql.Open("pgx", dataSourceName)
+	u.Path = "/postgres"
+	db, err := sql.Open("pgx", u.String())
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


brandond:
Can you rebase your changes? It appears that you are missing some changes to the tests and this is causing CI to fail.

I suspect that this was broken by https://github.com/k3s-io/kine/pull/179 - we should add or modify the tests so that they also exercise the database creation code. I believe they all currently use the default database.

oz123:
@brandond thanks for reviewing my PR so quickly. I rebased the patch against the current master. :crossed_fingers: that the tests work now.

dereknola (approved):

PR #258, opened 2023-12-12
Attempt to continue when encountering errors in createDBIfNotExist

mortenlj:
The database might already exist even if we encounter errors in createDBIfNotExist, so instead of erroring out, log warnings and attempt to continue.

Fixes https://github.com/k3s-io/k3s/issues/9033

### DIFF
--- a/pkg/drivers/pgsql/pgsql.go
+++ b/pkg/drivers/pgsql/pgsql.go
@@ -137,7 +137,7 @@ func createDBIfNotExist(dataSourceName string) error {
 	u.Path = "/postgres"
 	db, err := sql.Open("pgx", u.String())
 	if err != nil {
-		logrus.Warnf("failed to ensure existence of database %s; unable to connect to database postgres: %v", dbName, err)
+		logrus.Warnf("failed to ensure existence of database %s: unable to connect to default postgres database: %v", dbName, err)
 		return nil
 	}
 	defer db.Close()
--- a/pkg/drivers/pgsql/pgsql.go
+++ b/pkg/drivers/pgsql/pgsql.go
@@ -135,7 +135,6 @@ func createDBIfNotExist(dataSourceName string) error {
 
 	dbName := strings.SplitN(u.Path, "/", 2)[1]
 	u.Path = "/postgres"
-	var exists bool
 	db, err := sql.Open("pgx", u.String())
 	if err != nil {
 		logrus.Warnf("failed to ensure existence of database %s; unable to connect to database postgres: %v", dbName, err)
@@ -143,6 +142,7 @@ func createDBIfNotExist(dataSourceName string) error {
 	}
 	defer db.Close()
 
+	var exists bool
 	err = db.QueryRow("SELECT 1 FROM pg_database WHERE datname = $1", dbName).Scan(&exists)
 	if err != nil && err != sql.ErrNoRows {
 		logrus.Warnf("failed to check existence of database %s, going to attempt create: %v", dbName, err)
--- a/pkg/drivers/pgsql/pgsql.go
+++ b/pkg/drivers/pgsql/pgsql.go
@@ -138,14 +138,14 @@ func createDBIfNotExist(dataSourceName string) error {
 	var exists bool
 	db, err := sql.Open("pgx", u.String())
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/drivers/pgsql/pgsql.go
```diff
@@ -135,16 +135,17 @@ func createDBIfNotExist(dataSourceName string) error {
...
-	defer db.Close()
+		logrus.Warnf("failed to check existence of database %s; unable to connect to postgres database: %v", dbName, err)
+	} else {
+		defer db.Close()
 
-	var exists bool
-	err = db.QueryRow("SELECT 1 FROM pg_database WHERE datname = $1", dbName).Scan(&exists)
-	if err != nil && err != sql.ErrNoRows {
-		return err
+		err = db.QueryRow("SELECT 1 FROM pg_database WHERE datname = $1", dbName).Scan(&exists)
```
brandond, line 145: can you scan an untyped int into a bool?

pkg/drivers/pgsql/pgsql.go
```diff
@@ -153,9 +154,10 @@ func createDBIfNotExist(dataSourceName string) error {
 		logrus.Tracef("SETUP EXEC : %v", util.Stripped(stmt))
 		_, err = db.Exec(stmt)
```
brandond, line 155: This leaves us trying to use `db`, but if we don't return above when the open fails, we're operating with nil `db`, which will crash when you try to access the Exec function on a nil pointer.

brandond (changes requested):
I am pretty sure this will crash on error as-is.

pkg/drivers/pgsql/pgsql.go
```diff
@@ -153,9 +154,10 @@ func createDBIfNotExist(dataSourceName string) error {
...
 		if err != nil {
-			return err
+			logrus.Warnf("failed to create database %s: %v", dbName, err)
+		} else {
+			logrus.Tracef("created database: %s", dbName)
 		}
-		logrus.Tracef("created database: %s", dbName)
 	}
 	return nil
```
brandond, line 162: if we never return any errors, should this function even have a return value?

mortenlj:
> I am pretty sure this will crash on error as-is.

You are entirely right, not sure how I missed that. Serves me right for trying to crank it out just before bedtime :slightly_smiling_face:.

I'll take another pass over this today if I get the time.

pkg/drivers/pgsql/pgsql.go
mortenlj: This is just an indentation change, so no change really. But yes, you can. From the docs for `Rows.Scan`:

> For scanning into *bool, the source may be true, false, 1, 0, or string inputs parseable by strconv.ParseBool.

pkg/drivers/pgsql/pgsql.go
mortenlj: We return an error on line 133 if we can't parse the `dataSourceName`. Although, with the current code preceding the call site, `createDBIfNotExist` can never be called with an invalid `dataSourceName`, so that error will never happen. But it's good to handle it anyway, in case the calling function changes.

pkg/drivers/pgsql/pgsql.go
mortenlj: With the return on line 141, this shouldn't be possible anymore.

mortenlj:
@brandond I think I'm happy with this now, anything you think should be changed?

pkg/drivers/pgsql/pgsql.go
```diff
@@ -137,14 +137,15 @@ func createDBIfNotExist(dataSourceName string) error {
 	u.Path = "/postgres"
 	db, err := sql.Open("pgx", u.String())
 	if err != nil {
-		return err
+		logrus.Warnf("failed to ensure existence of database %s; unable to connect to database postgres: %v", dbName, err)
```
brandond, line 140:
```suggestion
		logrus.Warnf("failed to ensure existence of database %s: unable to connect to default postgres database: %v", dbName, err)
```

brandond (changes requested):
one nit on the log message, lgtm otherwise!

brandond (approved):

PR #264, opened 2024-01-08
check mysql db and table exist

yushiqie:
check mysql db and table exist before creating, because mysql user granted select,insert,update,delete on table kine can not create ddl sql in some case.

### DIFF
--- a/pkg/drivers/mysql/mysql.go
+++ b/pkg/drivers/mysql/mysql.go
@@ -7,14 +7,15 @@ import (
 	"fmt"
 
 	"github.com/go-sql-driver/mysql"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sirupsen/logrus"
+
 	"github.com/k3s-io/kine/pkg/drivers/generic"
 	"github.com/k3s-io/kine/pkg/logstructured"
 	"github.com/k3s-io/kine/pkg/logstructured/sqllog"
 	"github.com/k3s-io/kine/pkg/server"
 	"github.com/k3s-io/kine/pkg/tls"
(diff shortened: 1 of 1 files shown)


brandond:
This looks good, thanks for the contribution!

Please sign-off your commits for the DCO check.

yushiqie:
> This looks good, thanks for the contribution!
> 
> Please add sign-off to your commits for the DCO check.

yushiqie:
Thank you for the reminder, I have signed off !

pkg/drivers/mysql/mysql.go
```diff
@@ -24,7 +25,7 @@ const (
 
 var (
 	schema = []string{
-		`CREATE TABLE IF NOT EXISTS kine
```
brandond, line 27: Could we perhaps leave the create table/database statements alone? I'm not sure if it's possible to not have access to the `information_schema` table, but if so leaving these statements as-is would allow for graceful fallback to the current behavior.

brandond (changes requested):

pkg/drivers/mysql/mysql.go
```diff
@@ -43,7 +44,7 @@ var (
 		`CREATE INDEX kine_prev_revision_index ON kine (prev_revision)`,
 		`CREATE UNIQUE INDEX kine_name_prev_revision_uindex ON kine (name, prev_revision)`,
 	}
-	createDB = "CREATE DATABASE IF NOT EXISTS "
```
brandond, line 46: Revert this as well please

brandond (approved):

dereknola (approved):

Issue #267, opened 2024-01-12
Kine does not handle `WatchRequest_ProgressRequest` message


brandond:
@brandond I found a way to reproduce the watch-list request easily. Firstly, use kubectl proxy to bypass the authorization/certification issue.
```bash
kubectl proxy &
```
Then, use curl to create a watch-list request to apiserver, and please ensure you enabled watch-list feature in k8s apiserver.
```bash
curl 'http://localhost:8001/api/v1/namespaces/default/pods?allowWatchBookmarks=true&resourceVersionMatch=NotOlderThan&sendInitialEvents=true&timeoutSeconds=565&watch=true'
```
Now You can dig out what's going on under the hood.

_Originally posted by @qwtsc in https://github.com/k3s-io/kine/issues/255#issuecomment-1888178259_

PR #268, opened 2024-01-16
Add support for WatchProgressRequest

brandond:
* Adds support for watch progress requests.
  If all channels are synced, sends the current revision using channel -1. This allows the watch client to determine the datastore's current revision on-demand in between progress report intervals.

  This matches the implementation from upstream etcd: 
  * https://github.com/etcd-io/etcd/blob/v3.5.11/server/mvcc/watchable_store.go#L485-L487
  * https://github.com/etcd-io/etcd/blob/v3.5.11/server/mvcc/watchable_store.go#L500-L504
  For:
  * https://github.com/k3s-io/kine/issues/267
* Make progress report notification interval configurable.
  Required by kubernetes as per https://github.com/kubernetes/kubernetes/issues/122805#issuecomment-1898386268
* Optimize CurrentRevision checks by sharing current revision marker from poll loop
  Progress notifications query the current revision frequently; this can be optimized by exposing the poll loop's current revision once the loop has been started.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	go.etcd.io/etcd/client/v3 v3.5.9
 	go.etcd.io/etcd/server/v3 v3.5.9
 	google.golang.org/grpc v1.56.3
+	k8s.io/apimachinery v0.25.4
 	k8s.io/client-go v0.25.4
 )
 
@@ -90,7 +91,6 @@ require (
 	google.golang.org/protobuf v1.30.0 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
-	k8s.io/apimachinery v0.25.4 // indirect
 	k8s.io/klog/v2 v2.70.1 // indirect
 	k8s.io/utils v0.0.0-20220728103510-ee6ede2d64ed // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
--- a/main.go
+++ b/main.go
@@ -96,6 +96,12 @@ func main() {
 			Usage:       "Enable net/http/pprof handlers on the metrics bind address. Default is false.",
 			Destination: &metricsConfig.EnableProfiling,
 		},
+		cli.DurationFlag{
+			Name:        "watch-progress-notify-interval",
+			Usage:       "Interval between periodic watch progress notifications. Default is 10m.",
+			Destination: &config.NotifyInterval,
+			Value:       time.Minute * 10,
+		},
 		cli.BoolFlag{Name: "debug"},
 	}
 	app.Action = run
--- a/pkg/endpoint/endpoint.go
+++ b/pkg/endpoint/endpoint.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"os"
 	"strings"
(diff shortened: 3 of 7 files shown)


brandond:
@qwtsc @onprem are you able to test this?

qwtsc:
> @qwtsc @onprem are you able to test this?

good! let me have a try!

qwtsc:
@brandond Bad news, it does not work in my e2e test environment.

brandond:
Are you able to run Kine in debug mode, and share the results?

qwtsc:
My e2e test doesn't work with real etcd as backend storage either. I am so confused now. @onprem can you share the settings of apiserver?

qwtsc:
FYI, the related [issue](https://github.com/kubernetes/kubernetes/issues/122805) has been reported in k&k.

brandond:
I commented on the Kubernetes issue you linked, but I think the behavior you're expecting was incorrect and is no longer observed by v1.27+

brandond:
Hmm. If I add info-level logging on progress requests in kine, I see that they are CONSTANTLY being spammed by the apiserver. I also note that things seem to take way longer to start up when the featuregate is enabled. This feels pretty broken at the moment, and I don't think it is etcd or kine - I think the feature does not work properly.

brandond:
@qwtsc @onprem are you able to test with the commit I just pushed? I am still seeing some weirdness from the apiserver, but I believe the implementation is correct.

qwtsc:
> @qwtsc @onprem are you able to test with the commit I just pushed? I am still seeing some weirdness from the apiserver, but I believe the implementation is correct.

Ok, Let's see if it works or not.

qwtsc:
@brandond unfortunately, it doesn't work, but same test case works for etcd with special flag open.

brandond:
`experimental-watch-progress-notify-interval` is not a Boolean flag; it requires a value and overrides the default 10 minute watch progress interval. What are you setting it to?

Is your test case just the previously discussed curl request, or do you have something else that you can share?

brandond:
Oh, I see from the upstream k/k issue that it must be set to 5 seconds instead of the default 10 minutes, for the function to work. I didn't see that documented anywhere; let me take a look at making that configurable in kine.

brandond:
OK, please try with the most recent commit, using the new CLI flag `--watch-progress-notify-interval=5s`

qwtsc:
> `experimental-watch-progress-notify-interval` is not a Boolean flag; it requires a value and overrides the default 10 minute watch progress interval. What are you setting it to?
> 
> Is your test case just the previously discussed curl request, or do you have something else that you can share?

Something like curl, but it's written in golang.

>  try with the most recent commit, using the new CLI flag --watch-progress-notify-interval=5s

let's try. BTW, you're so hard working, I've noticed progress in this PR almost every six hours. Do you ever take time to rest?:-)

brandond:
> Do you ever take time to rest? :-)

We have had a weeklong ice storm and my kids are out of school, so I don't have much else to do lol

qwtsc:
Cheers, now It works smoothly. Thanks for your impressive contribution. @brandond

dereknola (approved):

vitorsavian (approved):

PR #271, opened 2024-01-25
fixes the incorrect total items count in a paginated list items q…

HemantNegi:
When making a list query the returned total matched items count is wrong.

**Steps to replicate:**
create some keys in the database with specific prefix say `/registry` then query with limit clause.

```
kine ❯ etcdctl get --prefix /registry --limit 1 --write-out=json | jq
{
  "header": {
    "revision": 494943
  },
  "kvs": [
    {
      "key": "L3JlZ2lzdHJ5L2hlYWx0aA==",
      "create_revision": 2,
      "mod_revision": 2,
      "value": "eyJoZWFsdGgiOiJ0cnVlIn0="
    }
  ],
  "more": true,
  "count": 2
}
```

the count shows wrong value.

**After this Fix**
```
kine ❯ etcdctl get --prefix /registry --limit 1 --write-out=json | jq
{
  "header": {
    "revision": 495036
  },
  "kvs": [
    {
      "key": "L3JlZ2lzdHJ5L2hlYWx0aA==",
      "create_revision": 2,
      "mod_revision": 2,
      "value": "eyJoZWFsdGgiOiJ0cnVlIn0="
    }
  ],
  "more": true,
  "count": 1842
}
```

### DIFF
--- a/pkg/logstructured/logstructured.go
+++ b/pkg/logstructured/logstructured.go
@@ -22,7 +22,7 @@ type Log interface {
 	List(ctx context.Context, prefix, startKey string, limit, revision int64, includeDeletes bool) (int64, []*server.Event, error)
 	After(ctx context.Context, prefix string, revision, limit int64) (int64, []*server.Event, error)
 	Watch(ctx context.Context, prefix string) <-chan []*server.Event
-	Count(ctx context.Context, prefix string) (int64, int64, error)
+	Count(ctx context.Context, prefix string, revision int64) (int64, int64, error)
 	Append(ctx context.Context, event *server.Event) (int64, error)
 	DbSize(ctx context.Context) (int64, error)
 }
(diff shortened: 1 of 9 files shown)


HemantNegi:
This problem was originally noticed while we were using Kine with a k8s api server. We make a paginated list objects api call and it returned incorrect `remainingItemCount`.

Please consider merging this or else fix this issue.

pkg/server/list.go
```diff
@@ -53,6 +53,14 @@ func (l *LimitedServer) list(ctx context.Context, r *etcdserverpb.RangeRequest)
 	if limit > 0 && resp.Count > r.Limit {
 		resp.More = true
 		resp.Kvs = kvs[0 : limit-1]
+
+		// count the actual number of results if there are more items in the db.
+		_, count, err := l.backend.Count(ctx, prefix)
```
brandond, line 58: I think this might need to be modified to take a `revision` that is set by the revision returned by the list call, otherwise you might get an inconsistent result. For example, if keys in this prefix are created or deleted in between the time the list completes, and when the count call is made.

brandond (changes requested):
Thanks for the PR! I do have one note about read consistency on the count call.

brandond:
Can also test this with: 
```bash
kubectl proxy &
curl -s 'http://localhost:8001/api/v1/namespaces/kube-system/pods?limit=1&resourceVersion=1000&resourceVersionMatch=Exact'
```

On a default k3s cluster, there should be 6 items remaining.

pkg/logstructured/logstructured.go
```diff
@@ -198,15 +198,16 @@ func (l *LogStructured) List(ctx context.Context, prefix, startKey string, limit
...
-		logrus.Tracef("COUNT %s => rev=%d, count=%d, err=%v", prefix, revRet, count, err)
+		logrus.Tracef("COUNT %s, rev=%d => rev=%d, count=%d, err=%v", prefix, revision, revRet, count, err)
 	}()
-	rev, count, err := l.log.Count(ctx, prefix)
+	rev, count, err := l.log.Count(ctx, prefix, revision)
 	if err != nil {
 		return 0, 0, err
 	}
 
+	// FIXME
```
dereknola, line 210: Context?

brandond: lol I did fix it, but I forgot to remove the note.

HemantNegi:
Thankyou for taking quick action on this and fixing other case as well. Looking forward to get it merged.

PR #273, opened 2024-02-06
Update schema and add migrations for mysql and pgsql

brandond:
Extends id column to bigint for new databases.

Existing databases will be migrated if the KINE_SCHEMA_MIGRATION environment variable is set to 1 or higher.

### DIFF
--- a/pkg/drivers/mysql/mysql.go
+++ b/pkg/drivers/mysql/mysql.go
@@ -5,6 +5,8 @@ import (
 	cryptotls "crypto/tls"
 	"database/sql"
 	"fmt"
+	"os"
+	"strconv"
 
 	"github.com/go-sql-driver/mysql"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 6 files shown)


vitorsavian (approved):

pkg/drivers/mysql/mysql.go
```diff
@@ -44,6 +46,9 @@ var (
 		`CREATE INDEX kine_prev_revision_index ON kine (prev_revision)`,
 		`CREATE UNIQUE INDEX kine_name_prev_revision_uindex ON kine (name, prev_revision)`,
 	}
+	schemaMigrations = []string{
+		`ALTER TABLE kine MODIFY COLUMN id BIGINT UNSIGNED NOT NULL UNIQUE, MODIFY COLUMN create_revision BIGINT UNSIGNED, MODIFY COLUMN prev_revision BITINT UNSIGNED`,
```
matttrach, line 50: Is 'AUTO_INCREMENT' assumed on 'NOT NULL' with 'UNIQUE'?

matttrach (approved):

pkg/drivers/mysql/mysql.go
brandond: My understanding is that when modifying a primary key column, you should explicitly include anything that's implied by  `PRIMARY KEY (id)` being present in the table create statement.

brandond: This statement has already been validated by Hosted Rancher as per https://github.com/k3s-io/k3s/issues/6176#issuecomment-1920386987

brandond: I'll add a test that starts up an old version of kine and then runs this migration, just to be sure.

galal-hussein (approved):

PR #278, opened 2024-02-24
Add GHCR image build

mjudeikis:
Adds no-cost image build on release to GHCR


dereknola:
This is now handles in GHA workflow

Issue #272, opened 2024-01-30
Document options for etcd client authentication in tcp listener


gberche-orange:
Thanks for sharing the kine component with the community. I'm looking for ways to secure a standalone kine cluster 

https://kubernetes.io/docs/tasks/administer-cluster/configure-upgrade-etcd/#limiting-access-of-etcd-clusters mentions
> After configuring secure communication, restrict the access of etcd cluster to only the Kubernetes API servers. Use TLS authentication to do so.
> 
> For example, consider key pairs k8sclient.key and k8sclient.cert that are trusted by the CA etcd.ca. When etcd is configured with --client-cert-auth along with TLS, it verifies the certificates from clients by using system CAs or the CA passed in by --trusted-ca-file flag. Specifying flags --client-cert-auth=true and --trusted-ca-file=etcd.ca will restrict the access to clients with the certificate k8sclient.cert.
> 
> Once etcd is configured correctly, only clients with valid certificates can access it. To give Kubernetes API servers the access, configure them with the flags --etcd-certfile=k8sclient.cert, --etcd-keyfile=k8sclient.key and --etcd-cafile=ca.cert.
> Note: etcd authentication is not currently supported by Kubernetes. For more information, see the related issue [Support Basic Auth for Etcd v2](https://github.com/kubernetes/kubernetes/issues/23398).
> 

However, from available cli options, I can't find the equivalent of `--client-cert-auth=true` in kine when accepting tcp connections

https://github.com/k3s-io/kine/blob/e0ae6642766a955cd4603bf749e010c7c689e0c9/main.go#L28-L99

Does kine authN and authZ instead only support unix socket (i.e.when no `--listen-address`flag is passed, and instead `unix://kine.sock` is read) by 3rd party component ?

brandond:
kine does not have any authn/authz whatsoever. Anyone that can connect to the etcd endpoint is allowed to access it. 

Configuring TLS on the listener (via the --server-cert-file/--server-key-file options) does not add any client certificate validation, it just enables use of a server certificate on the listening port.

The upstream kubernetes and etcd docs do not generally apply to kine; kine implements just enough of the etcd API to be usable by Kubernetes and additional bits of functionality like RBAC are optional and therefore not implemented.

I can take a look at enabling client certificate auth, but given that our most common deployment scenarios see kine being accessed via a unix socket where filesystem permissions can be used to restrict access, it may not be a high priority.

gberche-orange:
[...]

PR #279, opened 2024-02-26
Add a return when kine is set to use tls

vitorsavian:
- Returns the config that will be used in the client in k3s in case we use tls

### DIFF
--- a/pkg/endpoint/endpoint.go
+++ b/pkg/endpoint/endpoint.go
@@ -107,7 +107,9 @@ func Listen(ctx context.Context, config Config) (ETCDConfig, error) {
 	return ETCDConfig{
 		LeaderElect: leaderelect,
 		Endpoints:   []string{endpoint},
-		TLSConfig:   tls.Config{},
+		TLSConfig: tls.Config{
+			CAFile: config.ServerTLSConfig.CAFile,
+		},
 	}, nil
 }
 


pkg/endpoint/endpoint.go
```diff
@@ -104,10 +104,19 @@ func Listen(ctx context.Context, config Config) (ETCDConfig, error) {
...
+			Endpoints:   []string{endpoint},
+			TLSConfig:   tls.Config{},
+		}, nil
+	}
+
 	return ETCDConfig{
 		LeaderElect: leaderelect,
 		Endpoints:   []string{endpoint},
-		TLSConfig:   tls.Config{},
+		TLSConfig:   config.ServerTLSConfig,
```
brandond, line 119: I'm not sure this is correct. The TLSConfig in this block is intended to be used by clients - it should contain the CA chain and **client** cert+key that clients will use when connecting to the server. The ServerTLSConfig contains the **server** cert+key that is used for the listener. 

The best thing to do would probably be to return a new `tls.Config` with the CA cert from the ServerTLSConfig in the trusted root cert list. The client does not need a valid client cert+key, as kine does not currently support client certificate auth (see https://github.com/k3s-io/kine/issues/272#issuecomment-1917767824) - that is probably something we should fix at some point.

brandond (changes requested):

pkg/endpoint/endpoint.go
```diff
@@ -104,10 +104,21 @@ func Listen(ctx context.Context, config Config) (ETCDConfig, error) {
...
 
+	// Return a empty tls config if we do not receive any tls config
+	if config.ServerTLSConfig == (Config{}.ServerTLSConfig) {
+		return ETCDConfig{
+			LeaderElect: leaderelect,
+			Endpoints:   []string{endpoint},
+			TLSConfig:   tls.Config{},
+		}, nil
+	}
+
```
brandond, line 115:
```suggestion
```

pkg/endpoint/endpoint.go
```diff
@@ -104,10 +104,21 @@ func Listen(ctx context.Context, config Config) (ETCDConfig, error) {
...
 	return ETCDConfig{
 		LeaderElect: leaderelect,
 		Endpoints:   []string{endpoint},
-		TLSConfig:   tls.Config{},
+		TLSConfig: tls.Config{
```
brandond, line 119: This bit should be all you need - since the `ServerTLSConfig` field is not a pointer, it cannot be nil. If `config.ServerTLSConfig.CAFile` is an empty string, then it's the same as not setting it at all - there's no point in having a separate block for it.

brandond (changes requested):

brandond (approved):

dereknola (approved):

PR #275, opened 2024-02-16
NATS: Fix illegal resource version from storage: 0 (#274)

jrovira-kumori:
Fix for #274.

I am guessing this issue did not appear with K3s because the first command it runs when initializing is a `CREATE /bootstrap/...`. Therefore future `COUNT ...` calls would always return `!= 0`.

This is not the case with Kubernetes KubeAPI which reads before writing anything and gets stuck in a loop of `illegal resource version from storage: 0`.

This PR fixes the issue with the NATS driver by incrementing the BucketRevision number to 1 by creating and deleting a bootstrap key if a BucketRevision of 0 is detected at startup.

Issue #274, opened 2024-02-15
Kubeadm + Kine + NATS possible bug: illegal resource version from storage: 0


jrovira-kumori:
I have a minimal reproduction of deploying a cluster with Kind (kubeadm) and replacing ETCD with Kine + NATS KV store. It works when Kine uses the default SQLite store (omitting the `--endpoint=...` flag).

Minimal repoduction:
https://github.com/jrovira-kumori/kine-nats-kv-bug-minimal-repro

Running the minimal repoduction you will find that the KubeAPI server is stuck in a loop emitting a high volume of logs stating `illegal resource version from storage: 0`.

The only issue that I have found with a similiar problem is #74 - when using CockroachDB. Not much help here. I have been seriously stuck with this for a while. Any clarification if it is still a WIP feature would be greatly appreciated.

brandond:
What version of kine are you using?

brandond:
cc @bruth

jrovira-kumori:
Thanks for the quick response! Software versions are as follows:

- Kubernetes `v1.29.2`
- Kine `v0.11.4`
- NATS `v2.10.5`

jrovira-kumori:
I have updated the minimal reproduction to make it much easier to follow. 

You can run k3s (with everything disabled **but** the KubeAPI component) and point it to the same Kine. After the initialization is done you can kill k3s and restart the Kubernetes KubeAPI and it now seems to work.

At least the stream of `illegal resource version from storage: 0` no longer appears.

The new minimal reproduction shows this.

jrovira-kumori:
Found the issue. Will publish a PR shortly.

PR #281, opened 2024-03-02
Re-enable direct gets and update nats-server and nats.go versions

bruth:


bruth:
@brandond I check out the Drone CI and noticed that for the NATS 2.10.11, the API server was failing to come up, but its not clear why: 
image
NATS 2.9.25 and NATS embedded both worked.

brandond:
@bruth as per https://github.com/k3s-io/kine/pull/275#issuecomment-1976094545 it sounds like this does not resolve the error?

bruth:
Checking it out again.

bruth:
Found the root cause.. I had disabled an internal API that the KV bucket normally uses a while back since it was buggy at the time. However that had the side effect of having certain requests time out since directs gets were disabled. The error reporting simply appears as a timeout since there is not receiver of that direct get API.

I validated the timeout no longer occurs when creating the health check key-value pair which will prevent the issue noted in #274. This has been likely affecting all deployments of kine+NATS.

Also confirmed the CI passed this time.

bruth:
I will update with an additional commit shortly to ensure that API is enabled for existing KV buckets.

bruth:
@brandond Assuming CI passes, this is now ready to be reviewed and merged.

brandond (approved):

Issue #230, opened 2023-10-26
Make 'compactMinRetain' and 'compactInterval' configurable


sukumargaonkar:
Usecase:
Planning to use kine+postgres to store mostly immutable data (very few updates though out its lifetime).  

Current scenario:
To keep db size low, kine supports compaction and cleans up old revisions as per below conditions (as described [here](https://github.com/k3s-io/kine/blob/c5f2d892d70778fb9ad15208d4ca8617c3c794da/pkg/logstructured/sqllog/sql.go#L102-L107))
- versions older than compactInterval (default compactInterval 5mins)
- Always retain the last 1000 revisions (default compactMinRetain is 1000)

The 'compactInterval' and 'compactMinRetain' both are constants and thus cannot be configures

Feature Request:
Make 'compactMinRetain' configurable. When dealing with rarely updating data, setting the 'compactMinRetain' value low would help keep db size low

PR #282, opened 2024-03-05
feat: compaction support

TylerGillson:
This PR adds support for gRPC compaction requests for all backends. For NATS, compaction is handled by Jetstream and the bucket's `History` setting, so compaction requests are no-ops.

Additionally, unused params were removed from the LimitedServer's `compact` and `create` methods.

Given that the default compaction interval of 5min is not exposed as a configuration option, it can be useful to be able to trigger ad hoc compactions, which partially addresses #230.

### Test code
```golang
package main

import (
	"context"
	"fmt"
	"os"
	"strconv"

	"go.etcd.io/etcd/etcdserver/etcdserverpb"
	"google.golang.org/grpc"
	"google.golang.org/grpc/credentials/insecure"
)

func main() {
	compactRevision := os.Args[1]
	cr, err := strconv.ParseInt(compactRevision, 10, 64)
	if err != nil {
		panic(err)
	}

	conn, err := grpc.Dial("localhost:2379",
		grpc.WithTransportCredentials(insecure.NewCredentials()),
	)
	if err != nil {
		panic(err)
	}
	defer conn.Close()
	fmt.Println("Connected to kine")

	r := &etcdserverpb.CompactionRequest{
		Revision: cr,
	}
	fmt.Println("Triggering compaction", "revision", r.Revision)

	c := etcdserverpb.NewKVClient(conn)
	resp, err := c.Compact(context.Background(), r)
	if err != nil {
		panic(err)
	}

	fmt.Println("Received compaction response with revision", resp.Header.Revision)
}
```

### Sample output
```
root@tyler-k11-8277bf2c:/home/demo# ./kine-compact 1755748
Connected to kine
Triggering compaction revision 1755748
Received compaction response with revision 1002
```

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -27,6 +27,7 @@ type Client interface {
 	Create(ctx context.Context, key string, value []byte) error
 	Update(ctx context.Context, key string, revision int64, value []byte) error
 	Delete(ctx context.Context, key string, revision int64) error
+	Compact(ctx context.Context, revision int64) (int64, error)
 	Close() error
 }
 
@@ -142,6 +143,11 @@ func (c *client) Delete(ctx context.Context, key string, revision int64) error {
 		return fmt.Errorf("revision %d doesnt match", revision)
 	}
 	return nil
+}
+
+func (c *client) Compact(ctx context.Context, revision int64) (int64, error) {
+	resp, err := c.c.Compact(ctx, revision)
+	return resp.Header.GetRevision(), err
 }
 
 func (c *client) Close() error {
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -147,7 +147,10 @@ func (c *client) Delete(ctx context.Context, key string, revision int64) error {
 
 func (c *client) Compact(ctx context.Context, revision int64) (int64, error) {
 	resp, err := c.c.Compact(ctx, revision)
-	return resp.Header.GetRevision(), err
+	if resp != nil {
+		return resp.Header.GetRevision(), err
+	}
+	return 0, err
 }
 
 func (c *client) Close() error {
--- a/pkg/drivers/nats/backend.go
+++ b/pkg/drivers/nats/backend.go
@@ -428,7 +428,7 @@ func (b *Backend) Watch(ctx context.Context, prefix string, startRevision int64)
 	}
 }
 
-// Compact is a no-op / not implemented.
+// Compact is a no-op / not implemented. Revision history is managed by the jetstream bucket.
 func (b *Backend) Compact(ctx context.Context, revision int64) (int64, error) {
 	return revision, nil
 }
--- a/pkg/drivers/nats/backend.go
+++ b/pkg/drivers/nats/backend.go
@@ -427,3 +427,8 @@ func (b *Backend) Watch(ctx context.Context, prefix string, startRevision int64)
 		CurrentRevision: rev,
 	}
 }
+
+// Compact is a no-op / not implemented.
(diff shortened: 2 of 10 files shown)


brandond:
If we enable this, then the built-in compaction loop doesn't serve much purpose. Can we perhaps put this behind a CLI flag / option, and disable the built-in compaction loop when client-initiated compaction is enabled?

TylerGillson:
> If we enable this, then the built-in compaction loop doesn't serve much purpose. Can we perhaps put this behind a CLI flag, and disable the built-in compaction loop when client-initiated compaction is enabled?

In my use case the default 5m interval is fine 99% of the time. When a certain situation arises, I trigger a couple of manual compactions over a shorter time period, then fall back to the default.

I'm totally open to adding a flag that disables the built-in compaction loop, but would you consider allowing both?

brandond:
upstream etcd has:
```
--auto-compaction-retention '0'
  Auto compaction retention length. 0 means disable auto compaction.
--auto-compaction-mode 'periodic'
  Interpret 'auto-compaction-retention' one of: periodic|revision. 'periodic' for duration based retention, defaulting to hours if no time unit is provided (e.g. '5m'). 'revision' for revision number based retention.
```

right now, kine's behavior is essentially as if it had those flags and it defaulted to `--auto-compaction-retention=5m --auto-compaction-mode=periodic`

Perhaps we could add those flags, and wire them up appropriately?

I will also note that kine will always refuse to compact the most recent 1000 revisions. That is hardcoded and I wouldn't recommend changing it, at the risk of having nodes miss rows from the database in the case of multiple nodes talking to the same sql db.

TylerGillson:
@brandond I'm happy to add support for those flags, but supposing they've been added, kine (and etcd) would respect ad hoc compaction requests, right? Other than in the edge case you mentioned.

If that's the case, what are your thoughts on merging this and adding the flags in a follow-up PR?

brandond:
So I just looked at this again and I see that you're adding support for the direct compact grpc method, without wiring in the compact transaction that the apiserver sends - the apiserver's automatic compaction is still a no-op. As long as we're not changing the interaction with the apiserver, I think we're OK.

brandond (approved):

pkg/client/client.go
```diff
@@ -144,6 +145,11 @@ func (c *client) Delete(ctx context.Context, key string, revision int64) error {
 	return nil
 }
 
+func (c *client) Compact(ctx context.Context, revision int64) (int64, error) {
+	resp, err := c.c.Compact(ctx, revision)
+	return resp.Header.GetRevision(), err
```
briandowns, line 150: If err isn't nil would we still want to call resp.Header...?

TylerGillson: The response should [always be non-nil](https://github.com/k3s-io/kine/blob/a9d2c9cfc272ed8e6c6dd88636511b922c7cd296/pkg/server/compact.go#L12). And `GetRevision` handles `Header` being nil, so long as the response itself is non-nil.

Would something like this be preferable?
```golang
resp, err := c.c.Compact(ctx, revision)
if resp != nil {
    return resp.Header.GetRevision(), err
}
return -1, err
```

brandond: Yeah, it should always be non-nil, but it doesn't hurt to check I guess. If it is nil, I think 0 is better to return than -1. That's an invalid revision and could cause other problems.

TylerGillson: Resolved in 074e33a3d3961b21f198d533386550fbeb47b7a9.

PR #316, opened 2024-08-21
Return an emulated etcd version in the status endpoint

### LINKED ISSUE #315: Support for Kubernetes 1.31
Kubernetes 1.31 has an advanced caching layer https://kubernetes.io/blog/2024/08/15/consistent-read-from-cache-beta/

This requires certain features to be implemented by etcd, specifically https://etcd.io/docs/v3.5/dev-guide/interacting_v3/#watch-progress

Creating this issue to check whether kine is compatible with this feature or not.

samj1912:
Fixes #315 

See https://github.com/kubernetes/kubernetes/blob/beb696c2c9467dbc44cbaf35c5a4a3daf0321db3/staging/src/k8s.io/apiserver/pkg/storage/feature/feature_support_checker.go#L157

Which expects that the etcd status endpoint return a version. This returns a default version allowing users to override it if necessary.

### DIFF
--- a/main.go
+++ b/main.go
@@ -98,15 +98,15 @@ func main() {
 		},
 		&cli.DurationFlag{
 			Name:        "watch-progress-notify-interval",
-			Usage:       "Interval between periodic watch progress notifications. Default is 10m.",
+			Usage:       "Interval between periodic watch progress notifications. Default is 5s to ensure support for watch progress notifications.",
 			Destination: &config.NotifyInterval,
-			Value:       time.Minute * 10,
+			Value:       time.Second * 5,
 		},
 		&cli.StringFlag{
 			Name:        "emulated-etcd-version",
-			Usage:       "The emulated etcd version to return on a call to the status endpoint. Defaults to 3.4.31 which is the least version that support Kubernetes watch progress feature.",
+			Usage:       "The emulated etcd version to return on a call to the status endpoint. Defaults to 3.5.13, in order to indicate support for watch progress notifications.",
 			Destination: &config.EmulatedETCDVersion,
-			Value:       "3.4.31",
+			Value:       "3.5.13",
 		},
 		&cli.BoolFlag{Name: "debug"},
 	}
--- a/main.go
+++ b/main.go
@@ -102,6 +102,12 @@ func main() {
 			Destination: &config.NotifyInterval,
 			Value:       time.Minute * 10,
 		},
+		&cli.StringFlag{
+			Name:        "emulated-etcd-version",
+			Usage:       "The emulated etcd version to return on a call to the status endpoint. Defaults to 3.4.31 which is the least version that support Kubernetes watch progress feature.",
+			Destination: &config.EmulatedETCDVersion,
+			Value:       "3.4.31",
+		},
 		&cli.BoolFlag{Name: "debug"},
 	}
 	app.Action = run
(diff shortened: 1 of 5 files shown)


pkg/server/server.go
```diff
@@ -11,16 +11,18 @@ import (
...
 
-func New(backend Backend, scheme string, notifyInterval time.Duration) *KVServerBridge {
+func New(backend Backend, scheme string, notifyInterval time.Duration, emulatedETCDVersion string) *KVServerBridge {
 	return &KVServerBridge{
 		limited: &LimitedServer{
 			notifyInterval: notifyInterval,
 			backend:        backend,
 			scheme:         scheme,
 		},
+		EmulatedETCDVersion: emulatedETCDVersion,
```
brandond, line 25:
```suggestion
	return &KVServerBridge{
		emulatedETCDVersion: emulatedETCDVersion,
		limited: &LimitedServer{
			notifyInterval: notifyInterval,
			backend:        backend,
			scheme:         scheme,
		},
```

pkg/server/server.go
```diff
@@ -11,16 +11,18 @@ import (
 )
 
 type KVServerBridge struct {
-	limited *LimitedServer
+	limited             *LimitedServer
+	EmulatedETCDVersion string
```
brandond, line 15: Nit: this isn't used outside this package, make it private
```suggestion
	emulatedETCDVersion string
	limited             *LimitedServer
```

main.go
```diff
@@ -102,6 +102,12 @@ func main() {
 			Destination: &config.NotifyInterval,
 			Value:       time.Minute * 10,
 		},
+		&cli.StringFlag{
+			Name:        "emulated-etcd-version",
+			Usage:       "The emulated etcd version to return on a call to the status endpoint. Defaults to 3.4.31 which is the least version that support Kubernetes watch progress feature.",
+			Destination: &config.EmulatedETCDVersion,
+			Value:       "3.4.31",
```
brandond, line 109: I'd prefer to claim to be 3.5.x, as that is the version of etcd code I referenced when adding WatchProgressNotify support to kine. 
```suggestion
			Usage:       "The emulated etcd version to return on a call to the status endpoint.  Defaults to 3.5.13, in order to indicate support for watch progress notifications.",
			Destination: &config.EmulatedETCDVersion,
			Value:       "3.5.13",
```

brandond (changes requested):
LGTM, with a few nits

An additional question: if we fix the version reporting, Kubernetes will try to use watch progress notifications. However, Kubernetes also requires you to set the `--experimental-watch-progress-notify-interval=5s` (instead of the default of `10m`) when using watch progress notifications. This is also true when using kine; it has the same default value for the `--watch-progress-notify-interval` flag as upstream etcd.

Would you mind modifying the markdown files in the examples folder to call this out, and show the flag being set?

samj1912:
@brandond would you be open to changing the default value to 5s?

brandond:
Yeah, I can't see how it would hurt anything. K3s already sets it to 5 seconds, for both etcd and kine, so that things work right. I'm kinda curious why etcd didn't change the default as well, given that kubernetes will behave oddly if you don't.

samj1912:
@brandond all comments should be addressed. Could you also release a new version with these changes once the PR is merged?

brandond (approved):

samj1912:
Note, I also added the following comment https://github.com/kubernetes/kubernetes/issues/122805#issuecomment-2302975431

Issue #319, opened 2024-09-03
Invalid endpoint is ignored and embedded sqlite is silently started


rmweir:
A non-empty endpoint of the incorrect format would cause an embedded sqlite backend to be used. By passing a non-empty endpoint the user is intending to use an external backend.

samj1912:
This can be closed now.

rmweir:
@samj1912 yep, thanks everyone! cc: @brandond @galal-hussein

PR #320, opened 2024-09-03
Reject invalid endpoint

rmweir:
**Problem:**
Prior, a non-empty endpoint of the incorrect format would cause an embedded sqlite backend to be used. By passing a non-empty endpoint the user is intending to use an external backend.

**Solution:**
Now, the endpoint will be validated and passing a non-empty endpoint of an incorrect format will result in an error.

**Issue:**
https://github.com/k3s-io/kine/issues/319


galal-hussein (approved):
LGTM

galal-hussein:
@rmweir Looks like the test is failing tho in testing sqlite:
```
[sqlite] #- Tail: /tmp/GSt1E0/kine/1/logs/system.log
[sqlite] time="2024-09-03T18:48:42.444315289Z" level=fatal msg="parsing storage endpoint: invalid storage endpoint [sqlite]. Non-empty storage endpoint should be of the format <network>://<address>"
```

rmweir:
@galal-hussein can we either change the test or explicity check for "sqlite"? Because the value gets totally discarded, its not evaluated.

rmweir:
@galal-hussein I added an explicit check in a new commit so you can review it. If you're okay with that I can squash.

pkg/endpoint/endpoint.go
```diff
@@ -257,3 +265,16 @@ func networkAndAddress(str string) (string, string) {
...
+
+// validateStorageEndpoint validates the given string can be used to infer backend and optionally extract connection
+// info
+func validateStorageEndpoint(str string) error {
+	if strings.Contains(str, "://") {
+		return nil
+	}
+	if str == "" || str == "sqlite" {
+		return nil
+	}
```
brandond, line 277: This feels like it's duplicating the check in the `switch network` block above. Can we keep the scheme checking in one place?

pkg/endpoint/endpoint.go
```diff
@@ -257,3 +265,16 @@ func networkAndAddress(str string) (string, string) {
...
+	// without this check kine will silently run embedded sqlite, despite user intending to use a passed backend
+	return fmt.Errorf("invalid storage endpoint [%s]. Non-empty storage endpoint should be of the format <network>://<address>", str)
```
brandond, line 279: 1. Don't log the DSN, as it may contain db credentials that we should not log
2. Lets call this the datastore endpoint in the message, instead of the storage endpoint, as that is the name of the CLI flag in K3s and RKE2.
3. Use standard component names as per https://datatracker.ietf.org/doc/html/rfc3986#section-3
```suggestion
	return errors.New("invalid datastore endpoint; endpoint should be a DSN URI in the format <scheme>://<authority>")
```

brandond (changes requested):

pkg/endpoint/endpoint.go
rmweir: @brandond if I move the validation to the switch under the network == "" case, I don't think this message makes sense. If network == "" then the string is clearly not of that format but we reserve two cases where it is acceptable.

brandond:
I would be willing to tighten this down to require a valid scheme. So if you want sqlite for example, you have to do at least `sqlite://` instead of just `sqlite`. I don't think we have documented anywhere that we support `sqlite`, and if you want to pass the path to the DB file you already have to use a proper URI, as seen at https://github.com/k3s-io/k3s/issues/8234#issuecomment-1693696281

Also, please do

rmweir:
@brandond I think passing an empty storage endpoint is how one usually uses sqlite. I think expecting either a valid DSN URI or empty to imply an embedded backend is required, makes enough sense. Can we go with that?

brandond:
yeah thats what I meant. Either leave it empty, or pass a URI with a valid scheme and separator. The URI can have an empty authority/path/parameters to get whatever the defaults are for that backend.

rmweir:
@brandond I just updated the latest commit to what I was thinking. Lmk if that's fine and I will squash. If you want me too add some option for "sqlite://" I can do that too but I think we're okay without it.

brandond:
1. looks like you need to sign off for DCO
2. I think you'll need to fix the sqlite test that was linked above: https://github.com/k3s-io/kine/blob/cb8c8747cb32de71c150265f721ff72d8cec6dcf/scripts/test-run-sqlite#L3

brandond (approved):
lgtm!

rmweir:
@brandond fixed test, also had to move validation to its own function as we use that network/address parsing function in multiple places where string could be empty, squashed all commits. Just take another pass on everything please as I don't want to sneak anything in.

scripts/test-run-sqlite
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 start-test() {
-    KINE_IMAGE=$IMAGE KINE_ENDPOINT="sqlite" provision-kine
+    KINE_IMAGE=$IMAGE KINE_ENDPOINT="" provision-kine
```
brandond, line 3: `[sqlite] time="2024-09-03T21:27:16Z" level=fatal msg="flag needs an argument: -endpoint"`

You can leave `--endpoint` unset, but you can't currently set it to an empty string... so you should either fix the CLI flag, or explicitly set it.

```suggestion
    KINE_IMAGE=$IMAGE KINE_ENDPOINT="sqlite://" provision-kine
```

brandond (changes requested):

scripts/test-run-sqlite
brandond: Actually I suspect that when the var is empty the script is just running `kine --endpoint` instead of `kine --endpoint ""` so the CLI framework is complaining that it doesn't have a value - which is valid.

brandond (approved):

rmweir:
@brandond @galal-hussein thanks for the fast replies! Will probably need someone to merge this for me as well once everything passes :grimacing:.

brandond:
I'll wait for @galal-hussein to review again, then either one of us can merge and tag!

PR #322, opened 2024-09-19
feat: support `dbName` with operators in name (minus, ...)

lucasfcnunes:
## Related

- https://github.com/loft-sh/vcluster/issues/2152

### DIFF
--- a/pkg/drivers/pgsql/pgsql.go
+++ b/pkg/drivers/pgsql/pgsql.go
@@ -165,7 +165,6 @@ func createDBIfNotExist(dataSourceName string) error {
 		logrus.Warnf("failed to check existence of database %s, going to attempt create: %v", dbName, err)
 	}
 
-	
 	if !exists {
 		stmt := fmt.Sprintf(createDB, dbName)
 		logrus.Tracef("SETUP EXEC : %v", util.Stripped(stmt))
--- a/pkg/drivers/pgsql/pgsql.go
+++ b/pkg/drivers/pgsql/pgsql.go
@@ -3,6 +3,7 @@ package pgsql
 import (
 	"context"
 	"database/sql"
+	"fmt"
 	"net/url"
 	"os"
 	"regexp"
--- a/pkg/drivers/pgsql/pgsql.go
+++ b/pkg/drivers/pgsql/pgsql.go
@@ -201,6 +201,10 @@ func prepareDSN(dataSourceName string, tlsInfo tls.Config) (string, error) {
 		u.Path = "/kubernetes"
 	}
 
+	// makes quoting database and schema reference the same unquoted identifier
+	// See: https://www.postgresql.org/docs/12/sql-syntax-lexical.html#:~:text=unquoted%20names%20are%20always%20folded%20to%20lower%20case
+	u.Path = strings.ToLower(u.Path)
+
 	queryMap, err := url.ParseQuery(u.RawQuery)
 	if err != nil {
 		return "", err
--- a/pkg/drivers/pgsql/pgsql.go
+++ b/pkg/drivers/pgsql/pgsql.go
@@ -50,7 +50,7 @@ var (
 	schemaMigrations = []string{
 		`ALTER TABLE kine ALTER COLUMN id SET DATA TYPE BIGINT, ALTER COLUMN create_revision SET DATA TYPE BIGINT, ALTER COLUMN prev_revision SET DATA TYPE BIGINT; ALTER SEQUENCE kine_id_seq AS BIGINT`,
 	}
-	createDB = "CREATE DATABASE "
+	createDB = `CREATE DATABASE "%s";`
 )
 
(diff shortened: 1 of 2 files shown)


lucasfcnunes:
Fixes loft-sh/vcluster#2152.

pkg/drivers/pgsql/pgsql.go
```diff
@@ -50,7 +50,7 @@ var (
 	schemaMigrations = []string{
 		`ALTER TABLE kine ALTER COLUMN id SET DATA TYPE BIGINT, ALTER COLUMN create_revision SET DATA TYPE BIGINT, ALTER COLUMN prev_revision SET DATA TYPE BIGINT; ALTER SEQUENCE kine_id_seq AS BIGINT`,
 	}
-	createDB = "CREATE DATABASE "
+	createDB = `CREATE DATABASE "%s";`
```
brandond, line 53: As discussed in an existing PR to fix this, quoting the database name makes it case-sensitive, which would be a breaking change for anyone that had previously been specifying a database name with uppercase characters, and was relying on the database engine to case-fold it to lowercase.

https://github.com/k3s-io/kine/pull/149#issuecomment-1330623666

Perhaps we could modify this to only quote the database name if it contains special characters, to preserve existing behavior?

brandond (changes requested):

pkg/drivers/pgsql/pgsql.go
lucasfcnunes: Wouldn't lowercasing `dbName` in such cases solve the issue?

lucasfcnunes: @brandond, WDYT?
https://github.com/k3s-io/kine/pull/322/commits/924c3ba971da83e4c8a230eb51fa10717112c508

brandond: Yeah, I dunno. What if someone needs to use mixed/uppercase characters alongside their special characters that need quoting? Maybe we should add an envelope var to control this; if set it unconditionally quotes, if not it only quotes it if it contains special characters?

lucasfcnunes: I think it would be very confusing to have the following.

- `postgres://.../kubernetes_dev` == `postgres://.../KUBERNETES_DEV` ✅ (uppercase turns into lowercase (because of missing quotes), no questions asked)
- `postgres://.../kubernetes-dev` != `postgres://.../KUBERNETES-DEV` ❌ (breaks, then you have to toggle a flag, but it does not work as the above)

Confusing code for confusing behavior.

brandond: Right, but there are two cases we need to consider:
1. User who has been using `postgres://.../MyDatabase` for a long time, and must not find this silently switched from `mydatabase` to `MyDatabase` after upgrade.
2. User who is just starting out, has no database, and wants to use `MyDatabase` as requested.

Perhaps we could unconditionally quote the name, but explicitly check for the lowercase variant first, and use that if present?

lucasfcnunes: > 1. User who has been using postgres://.../MyDatabase for a long time, and must not find this silently switched from mydatabase to MyDatabase after upgrade.

This is not gonna happen. See https://github.com/k3s-io/kine/pull/322/commits/924c3ba971da83e4c8a230eb51fa10717112c508.

> 2. User who is just starting out, has no database, and wants to use MyDatabase as requested.

This PR makes no changes in behaviors related to case-sensitiveness. All identifiers will keep being case folded to lowercase as always like in the unquoted case.
In my opinion, the possibility to turn on case-sensitive behavior should be done in another PR.

--

After the PR the expected behaviors in postgress will be as follows (because of https://github.com/k3s-io/kine/pull/322/commits/924c3ba971da83e4c8a230eb51fa10717112c508):
- For '_'
  - `postgres://.../kubernetes_dev` -> `CREATE DATABASE "kubernetes_dev"` (the equivalent to what has always been `CREATE DATABASE kubernetes_dev`)
  - `postgres://.../KUBERNETES_DEV` -> `CREATE DATABASE "kubernetes_dev"` (the equivalent to what has always been `CREATE DATABASE KUBERNETES_DEV`)
  -  `postgres://.../Kubernetes_Dev` -> `CREATE DATABASE "kubernetes_dev"` (the equivalent to what has always been `CREATE DATABASE Kubernetes_Dev`)
- For '-'
  - `postgres://.../kubernetes-dev` -> `CREATE DATABASE "kubernetes-dev"` (no prior working equivalent)
  - `postgres://.../KUBERNETES-DEV` -> `CREATE DATABASE "kubernetes-dev"` (no prior working equivalent)
  - `postgres://.../Kubernetes-Dev` -> `CREATE DATABASE "kubernetes-dev"` (no prior working equivalent)

lucasfcnunes: It would be great if I could test these cases in the CI.

lucasfcnunes:
@brandond, this PR fixes another issue https://github.com/k3s-io/kine/issues/324.

brandond (approved):

lucasfcnunes:
Fixes #148

Issue #76, opened 2021-02-13
Consider recognize `postgresql://` network protocol


Dentrax:
After a few hours of debugging `storage backend is not defined` error, we have just found that `kine` currently does not recognize `postgresql://` protocol , but does `postgres://`.

correct: `postgres://admin:admin@k3s.db.ondigitalocean.com:5432/defaultdb`

incorrect: `postgresql://admin:admin@k3s.db.ondigitalocean.com:5432/defaultdb`

PR #325, opened 2024-09-25
Refactor driver selection to allow registering new drivers

samj1912:
This PR refactors the driver code to allow users to register their own drivers if they wish to.

It also consolidates the I/O that is expected from each driver constructor to a consistent interface.

Fixes https://github.com/k3s-io/kine/issues/76 as well


pkg/drivers/generic/generic.go
```diff
@@ -25,6 +26,40 @@ const (
...
+type Config struct {
+	MetricsRegisterer    prometheus.Registerer
+	Endpoint             string
+	Scheme               string
+	DataSourceName       string
+	ConnectionPoolConfig ConnectionPoolConfig
+	BackendTLSConfig     tls.Config
+}
+
+type Constructor func(ctx context.Context, cfg *Config) (leaderElect bool, backend server.Backend, err error)
```
brandond, line 38: Might the registry better fit in `pkg/drivers`? This would make the calls in other packages a little more idiomatic - `drivers.Register()`, `drivers.SetDefault()/GetDefault()`, `drivers.Resolve()` or perhaps `drivers.ForScheme()` etc.

pkg/drivers/generic/generic.go
```diff
@@ -25,6 +26,40 @@ const (
...
+
+var driverRegistry map[string]Constructor = map[string]Constructor{}
+var defaultDriver string
+
+func RegisterDriver(driver string, constructor Constructor) {
+	driverRegistry[driver] = constructor
+}
+
+func SetDefaultDriver(driver string) {
```
brandond, line 47: Please add doc comments as well, since this is going to be an explicit public extension point.

```suggestion
func RegisterDriver(scheme string, constructor Constructor) {
	driverRegistry[scheme] = constructor
}

func SetDefaultDriver(scheme string) {
```

pkg/drivers/registry.go
```diff
@@ -0,0 +1,41 @@
...
+// SetDefault sets the default driver
+// The default driver is used when the endpoint is not specified
+func SetDefault(driver string) {
+	defaultDriver = driver
+}
+
+// GetDefault returns the default driver
+func GetDefault() Constructor {
+	return driverRegistry[defaultDriver]
+}
```
brandond, line 31:
```suggestion
var defaultScheme string

// Register registers a constructor for the given scheme
func Register(scheme string, constructor Constructor) {
	driverRegistry[scheme] = constructor
}

// SetDefault sets the default driver scheme
// The default driver is used when an endpoint is not specified
func SetDefault(scheme string) {
	defaultScheme = scheme
}

// GetDefault returns the default driver
func GetDefault() Constructor {
	return driverRegistry[defaultScheme]
}
```

brandond (changes requested):

pkg/drivers/registry.go
```diff
@@ -0,0 +1,39 @@
...
+
+// Get returns the constructor for the given scheme
+// The second return value is true if the scheme is registered, false otherwise
+func Get(scheme string) (Constructor, bool) {
+	constructor, ok := driverRegistry[scheme]
+	if !ok {
+		return nil, false
+	}
+	return constructor, true
```
brandond, line 38: I wish go was smart enough to use the two-value form if we did `return driverRegistry[scheme]` instead of needing to assign intermediate variables, but such is life. Regardless, you can just do:
```suggestion
	return constructor, ok
```

brandond (changes requested):
another nit

pkg/drivers/registry.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"context"
+
+	"github.com/k3s-io/kine/pkg/server"
+)
+
+// Constructor is a function that takes a context and a config and returns a leaderElect bool, a server.Backend and an error
+type Constructor func(ctx context.Context, cfg *Config) (leaderElect bool, backend server.Backend, err error)
+
+var driverRegistry map[string]Constructor = map[string]Constructor{}
```
brandond, line 12: revive lint is failing on this:
`var-declaration: should omit type map[string]Constructor from declaration of var driverRegistry; it will be inferred from the right-hand side (revive)`
```suggestion
var driverRegistry = map[string]Constructor{}
```

brandond (changes requested):

brandond:
Looks like it's still not formatted properly:
```
#12 1.968 pkg/drivers/generic/generic.go:28: File is not `gofmt`-ed with `-s` (gofmt)
```

samj1912:
Should work now, rebased and squashed all the fixes.

samj1912:
Hmm

> msg="building kine: no default driver found"

brandond (approved):

vitorsavian (approved):

samj1912:
@brandond ok to merge this?

PR #333, opened 2024-09-30
ci: increase open PRs for gomod on dependabot

lucasfcnunes:
The default limit is 5. Increasing it to 15 to have more visibility about Go packages upgradability.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -32,6 +32,7 @@ updates:
 
   # Maintain dependencies for Go
   - package-ecosystem: "gomod"
+    open-pull-requests-limit: 15
     directory: "/"
     labels:
       - "kind/dependabot"


brandond:
I'm not sure this will make a HUGE difference, as there should not in general be that many packages with pending updates - there's just a bit of a backlog at the moment.

brandond:
Ugh this is a train wreck, it's trying to bump all the modules individually instead of in groups - it's doing one bump for each of the many etcd modules, for example, and there are 4 of them that should be bumped together.

lucasfcnunes:
@brandond I think dependabot has a beta feature that does exactly this. But just for [security updates](https://github.blog/changelog/2023-12-07-dependabot-grouped-security-updates-public-beta/#:~:text=Starting%20today%2C%20you%20can%20enable%20grouped%20security%20updates,with%20all%20of%20them%2C%20per%20ecosystem%2C%20across%20directories.).

brandond:
Yeah... until then I guess we can just close out the duplicates and manually rework one of them to update the whole group.

brandond:
Oh no, we can configure grouping for version updates
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

Issue #363, opened 2024-11-09
Compact reports deleting more rows than revisions


brandond:
After running a load test against the cluster, compaction reports deleting more rows than revisions. I'm not sure if this is a bug or what, but it merits investigation.

```
INFO[10800] COMPACT compactRev=570142 targetCompactRev=571142 currentRev=575816
INFO[10800] COMPACT deleted 1967 rows from 1000 revisions in 64.858348ms - compacted to 571142/575816
INFO[10800] COMPACT compactRev=571142 targetCompactRev=572142 currentRev=575817
INFO[10800] COMPACT deleted 1965 rows from 1000 revisions in 64.937484ms - compacted to 572142/575817
INFO[10800] COMPACT compactRev=572142 targetCompactRev=573142 currentRev=575818
INFO[10800] COMPACT deleted 1968 rows from 1000 revisions in 61.562873ms - compacted to 573142/575818
INFO[10800] COMPACT compactRev=573142 targetCompactRev=574142 currentRev=575819
INFO[10800] COMPACT deleted 1969 rows from 1000 revisions in 395.45276ms - compacted to 574142/575819
INFO[10800] COMPACT compactRev=574142 targetCompactRev=574824 currentRev=575824
INFO[10800] COMPACT deleted 1343 rows from 682 revisions in 41.083903ms - compacted to 574824/575824
INFO[10800] COMPACT compactRev=574824 targetCompactRev=574825 currentRev=575825
INFO[10800] COMPACT deleted 2 rows from 1 revisions in 81.61197ms - compacted to 574825/575825
INFO[10800] COMPACT compactRev=574825 targetCompactRev=574828 currentRev=575828
INFO[10800] COMPACT deleted 6 rows from 3 revisions in 10.927536ms - compacted to 574828/575828
INFO[10800] COMPACT revision 574828 has already been compacted

INFO[11100] COMPACT compactRev=574828 targetCompactRev=575828 currentRev=577945
INFO[11100] COMPACT deleted 1970 rows from 1000 revisions in 144.060248ms - compacted to 575828/577945
INFO[11100] COMPACT compactRev=575828 targetCompactRev=576828 currentRev=577945
INFO[11100] COMPACT deleted 1978 rows from 1000 revisions in 120.049635ms - compacted to 576828/577945
INFO[11100] COMPACT compactRev=576828 targetCompactRev=576945 currentRev=577945
[...]

PR #365, opened 2024-11-11
Fix rolling compact

brandond:
Fixes https://github.com/k3s-io/kine/issues/363

Undo the change from #360 and fix things the right way. That worked, but had the unfortunate side effect of moving the compact target forward within the inner loop. This whole thing can be done better by simplifying the error handling.

* Revert "Record compact progress and metrics per iteration"
    This reverts commit aa4f2a7c621b64aba27c6b1298c0481e0c0018bf.
* Simplify compact loop error handling
    Rather than restarting the outer loop on unhandled error, simmply break out of the inner loop on any error. We always want to update the compact and target rev, and do post-compact operations - so the only thing we really need to check the error for is logging and metrics.

### DIFF
--- a/pkg/logstructured/sqllog/sql.go
+++ b/pkg/logstructured/sqllog/sql.go
@@ -113,7 +113,6 @@ func (s *SQLLog) compactor(interval time.Duration) {
 	targetCompactRev, _ := s.CurrentRevision(s.ctx)
 	logrus.Tracef("COMPACT starting compactRev=%d targetCompactRev=%d", compactRev, targetCompactRev)
 
-outer:
 	for {
 		select {
 		case <-s.ctx.Done():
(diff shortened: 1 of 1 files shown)


dereknola (approved):

dereknola (approved):

cagataygurturk:
Hi, we've upgraded from 1.13.2 to 1.13.7 and we've encountered a huge performance problem with the PostgreSQL server. For full context, we are running 30+ clusters with the same SQL server.

After the upgrade, disk space usage started to increase, query times hugely increased and the SQL server came to an halt. Once we downgraded, we've seen a tons of records being deleted and eventually the performance came back to original state.

The only relevant change between these two versions is this, that's why I am commenting here. Any help?

brandond:
https://github.com/k3s-io/kine/compare/v0.13.2...v0.13.7

The change here doesn't impact the compact query itself. It only changes when the revision is recorded. Can you provide compact time metrics or log output from before and after the upgrade? As well as some information on your db? What version, how many rows, total size, number of connected servers, etc.

PR #366, opened 2024-11-12
fix: close mysql connection after create database

FabianKramm:
One problem we encountered when using Kine with the MySQL driver is that it uses an unexpected high amount of database connections, even after setting the flag `--datastore-max-open-connections` to a small number. 

We investigated and found out that during the `createDBIfNotExist` function two database connections are created as well that are never closed. For the PostgreSQL driver the connection is closed as expected:
https://github.com/k3s-io/kine/blob/133503d8acbf5979279e32b5a1ff28882c70212a/pkg/drivers/pgsql/pgsql.go#L223

For the MySQL driver, these close after creation defers are missing and hence the connections stay open all the time. This PR adds two lines to close the database connections as expected after making sure the database exists in the MySQL server.

EDIT: Not sure why the drone ci build is failing, but it seems unrelated to this change

### DIFF
--- a/pkg/drivers/generic/generic.go
+++ b/pkg/drivers/generic/generic.go
@@ -406,6 +406,7 @@ func (d *Generic) IsFill(key string) bool {
 	return strings.HasPrefix(key, "gap-")
 }
 
+//nolint:revive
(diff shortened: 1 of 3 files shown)


brandond (approved):

FabianKramm:
@brandond I ignored the linter for the 2 functions, tests seem to run now

dereknola (approved):

PR #373, opened 2024-12-04
fix: canceled should always be set to true when cancel a watch request

nic-6443:
Anytime the server attempts to cancel a watch request, it should set the `canceled` field to true, regardless of whether `err` is empty or not. Let `err` only affect the `reason` field.

### DIFF
--- a/pkg/logstructured/logstructured.go
+++ b/pkg/logstructured/logstructured.go
@@ -437,7 +437,7 @@ func (l *LogStructured) Watch(ctx context.Context, prefix string, revision int64
 				wr.CompactRevision = compact
 				wr.CurrentRevision = rev
 			} else {
-				errc <- errors.New("failed to execute query in database")
+				errc <- server.ErrGRPCUnhealthy
 			}
 		}
 		cancel()
--- a/pkg/logstructured/logstructured.go
+++ b/pkg/logstructured/logstructured.go
@@ -425,17 +425,20 @@ func (l *LogStructured) Watch(ctx context.Context, prefix string, revision int64
 	}
 
 	result := make(chan []*server.Event, 100)
-	wr := server.WatchResult{Events: result}
+	errc := make(chan error, 1)
+	wr := server.WatchResult{Events: result, Errorc: errc}
 
 	rev, kvs, err := l.log.After(ctx, prefix, revision, 0)
 	if err != nil {
 		if !errors.Is(err, context.Canceled) {
 			logrus.Errorf("Failed to list %s for revision %d: %v", prefix, revision, err)
-		}
-		if err == server.ErrCompacted {
-			compact, _ := l.log.CompactRevision(ctx)
-			wr.CompactRevision = compact
-			wr.CurrentRevision = rev
+			if err == server.ErrCompacted {
+				compact, _ := l.log.CompactRevision(ctx)
+				wr.CompactRevision = compact
+				wr.CurrentRevision = rev
+			} else {
+				errc <- errors.New("failed to execute query in database")
+			}
 		}
 		cancel()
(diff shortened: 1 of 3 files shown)


brandond:
Thanks! Can you provide any information on what issue this is causing so that we can validate the fix?

nic-6443:
> Thanks! Can you provide any information on what issue this is causing so that we can validate the fix?

Sure, the problem I encountered is as follows: I used kine as a replacement for etcd in the https://github.com/apache/apisix project, using postgres as the backend database. When performing maintenance operations on postgres, it needs to be restarted. If kine happens to receive a watch request at this time, this request will be canceled by kine due to SQL execution failure: https://github.com/k3s-io/kine/blob/73df6c7dfd811b3e3ec04b57174d7251912e29ef/pkg/logstructured/logstructured.go#L430 error log:
```
time="2024-09-24T20:20:23Z" level=error msg="Failed to list / for revision 3095: FATAL: terminating connection due to administrator command (SQLSTATE 57P01)"
```
Although kine cancels this request, it does not set the `canceled` field to true, causing abnormal client behavior. The response returned by kine is:
```
{"result":{"header":{},"watch_id":"1"}}
```
I have checked the native etcd code and found that when canceling a watch request, they always set `canceled` to true. I think we should stay consistent with etcd.

brandond:
That is interesting. This suggests that the error is not being propagated up from the failed list call, since `err != nil` is false and the cancelReason is empty. Are you able to run kine with --debug and find the output of the `WATCH CANCEL id=` log when this happens?

nic-6443:
> That is interesting. This suggests that the error is not being propagated up from the failed list call, since `err != nil` is false and the cancelReason is empty. Are you able to run kine with --debug and find the output of the `WATCH CANCEL id=` log when this happens?

this is the log:
```
TRAC[0022] WATCH CANCEL id=1, reason=, compactRev=0
```
I have considered whether to return the cause of the error to the client, such as `FATAL: terminating connection due to administrator command (SQLSTATE 57P01)`, but considering that this is internal server information, it seems inappropriate to return it to the client. If you think it is necessary to pass this error message through to the client, I can consider recording the error in `server.WatchResult` and returning it in the cancel response.

brandond:
I don't know if we need to expose the raw error message to the client, but it does seem correct to return AN error.

nic-6443:
@brandond send a error message to client now.

pkg/logstructured/logstructured.go
```diff
@@ -425,17 +425,20 @@ func (l *LogStructured) Watch(ctx context.Context, prefix string, revision int64
...
-		if err == server.ErrCompacted {
-			compact, _ := l.log.CompactRevision(ctx)
-			wr.CompactRevision = compact
-			wr.CurrentRevision = rev
+			if err == server.ErrCompacted {
+				compact, _ := l.log.CompactRevision(ctx)
+				wr.CompactRevision = compact
+				wr.CurrentRevision = rev
+			} else {
+				errc <- errors.New("failed to execute query in database")
```
brandond, line 440: nit: Add `ErrWatchCancelled = rpctypes.ErrGRPCWatchCanceled` in `pkg/server/types.go` and just return that here as a generic watch failure message. We try to return the same errors as upstream whenever possible, even if the cause isn't necessarily the same.

brandond (changes requested):
lgtm, one nit!

pkg/logstructured/logstructured.go
nic-6443: oh, great, I think `ErrGRPCUnhealthy` is better to express the failure come from system level.

brandond (approved):

PR #379, opened 2024-12-12
Update all testing DBs, remove EOL versions

dereknola:
- Update all the various DB we test, remove EOL versions.
- Reduce number of cockroachdb versions tested from 3 to 2. This DB takes significantly longer than other DBs to test.


brandond (approved):

scripts/test-run-mysql
```diff
@@ -23,11 +23,11 @@ start-test() {
 export -f start-test
 
 VERSION_LIST="\
+    mysql 8.4
     mysql 8.0
-    mysql 5.7
+    mariadb 11.4
```
brandond, line 28: looks like the connection test doesn't work for 11.4, I suspect something's changed in the CLI or password env var stuff.

brandond (approved):

vitorsavian (approved):

PR #392, opened 2025-01-10
Do not log context cancelled as error

mhmxs:
In my perspective context cancellation is not an error event, it is business as usual. WDYT?

### DIFF
--- a/pkg/server/kv.go
+++ b/pkg/server/kv.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"errors"
 
 	"github.com/sirupsen/logrus"
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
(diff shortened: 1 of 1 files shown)


brandond:
There are a LOT of places where functions that accept a Context may return an error. Other than a handful of specific places that caused a lot of log spam, we don't hide the error log if it is due to context cancellation. How did you single these out as places where we should suppress the error? Are you planning on adding any others?

mhmxs:
@brandond 
> How did you single these out as places where we should suppress the error? 

I ran 300h of various kubernetes e2e tests in the last two month and these were very annoying.

> Are you planning on adding any others?

I didn't plan, because all `sig-api-machinery|sig-apps|sig-auth|sig-instrumentation|sig-scheduling` tests are running without a single context cancellation error message, so i'm happy, but what do you suggest?

brandond (approved):

PR #393, opened 2025-01-10
Missing error handler at server list

mhmxs:
I think if list has an error it doesn't make any sense to do the rest of the function.


brandond:
If err is non-nil then the `resp.Count > r.Limit` test will be false (as resp.Count will be 0) and the block will not be executed.

Does this fix anything or are you just looking for things to change?

mhmxs:
@brandond i can agree with your view, in my personal taste exact error handling makes changing the code more easier.

PR #394, opened 2025-01-10
Git ignore debug files

mhmxs:
I think it makes things easier if temporary debug related builds are ignored in git.


mhmxs:
> What are you using that creates these files? I've never seen anything leave these in the project directory.

vscode

Issue #374, opened 2024-12-06
Missing check for NullInt64.Valid in Generic.Count


indradhanush:
We're running kine with mysql and see this error intermittently:

```
│ time="2024-12-04T18:12:50.283533302Z" level=error msg="error while range on /registry/statefulsets/ /registry/statefulsets/: sql: Scan error on column index 0, name \"prev_revision\": converting NULL to int64 is unsupported"
```

The error originates when [LimitedServer.Range invokes the method `list`](https://github.com/k3s-io/kine/blob/c8e68ac5a5d7cea4959aad7d66200d50cbbd06d1/pkg/server/limited.go#L20) and ultimately invoking [Generic.Count](https://github.com/k3s-io/kine/blob/c8e68ac5a5d7cea4959aad7d66200d50cbbd06d1/pkg/drivers/generic/generic.go#L379).

While the error goes away when we restart kine (thus our hunch being this related to a race condition) but it does show up again after a few days. We haven't yet narrowed down on what triggers this. 

In the meantime we would be interested in contributing a fix for this. While the fix in itself is simple - check for `NullInt64.Valid`, but we're not sure what the error handling should do. Should it return a custom error? I'd love your thoughts before I work on a PR.

Note: The same handling should also be applied to `Generic.CountCurrent` and perhaps also to the other usage of `NullInt64` in [sqllog.scan](https://github.com/k3s-io/kine/blob/c8e68ac5a5d7cea4959aad7d66200d50cbbd06d1/pkg/logstructured/sqllog/sql.go#L600).

brandond:
What version of mysql are you using? Can you reproduce this on demand?

indradhanush:
We're running MySQL 8.0.25. Unfortunately we haven't been able to reliably reproduce this yet.


> I don't think it's in generic.Count because that only uses the max revision via SELECT MAX(rkv.id) AS id FROM kine AS rkv, and the error is about the prev_revision column which is only needed when reading an entry.

I traced the usage of that query which is defined as `compactRevSQL` and see that:

[...]

PR #395, opened 2025-01-10
Fix compact rev sql on case of missing

mhmxs:
There is a very little chance that `compactRevSQL` is returning an empty result, which breaks other query executions. Coalesce ensures the query always returns a number.


brandond:
There is some discussion about this at https://github.com/k3s-io/kine/issues/374. The compact_rev_key row should never be absent. How are you able to reproduce this state? Even if you manually delete that row with sqlite CLI, it will be recreated on startup.

PR #396, opened 2025-01-10
Fix watch event on case of delete

mhmxs:
This change targets two small but significant issues. I discovered this by running wide variety of Kubernetes e2e tests and dag into cache how it works.
- Kubernetes doesn't like `nil` `Kv` or `PreKv`, except for deletion where it is mandatory to be `nil`
- In case of deletion it is looking for `PrevKv`, but log structured sends `Kv`. I know it is possible to fix this at `logstructured.go` but in this way Kine should guarantee the right behavior without depending on plugin correctness.

WDYT?

### DIFF
--- a/pkg/server/kv.go
+++ b/pkg/server/kv.go
@@ -71,7 +71,7 @@ func toKVs(kvs ...*KeyValue) []*mvccpb.KeyValue {
 
 	ret := make([]*mvccpb.KeyValue, 0, len(kvs))
 	for _, kv := range kvs {
-		newKV := toKV(kv)
+		newKV := toKV(kv, nil)
 		if newKV != nil {
 			ret = append(ret, newKV)
 		}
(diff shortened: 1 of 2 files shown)


brandond:
Can you open an issue describing what specific issues you saw and how to reproduce them?

mhmxs:
> Can you open an issue describing what specific issues you saw and how to reproduce them?

@brandond sorry my memories are ... so the problem is not in `logstructured.go`.

I created a driver implementation based on https://github.com/k3s-io/kine/blob/edd8f35371395a3b0c4ea229061f46807855bed7/pkg/drivers/nats/backend.go#L52 

kubernetes e2e tests failed, and after some debugging, i found this:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go#L693

So current object is not necessary, old object is mandatory, otherwise it fails to detect event type here:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go#L626

Two things came into my mind:
- save the planet by not sending current object like `nats/backend.go` and possibly `logstructured.go` do
  - https://github.com/k3s-io/kine/blob/edd8f35371395a3b0c4ea229061f46807855bed7/pkg/drivers/nats/backend.go#L409
  - https://github.com/k3s-io/kine/blob/edd8f35371395a3b0c4ea229061f46807855bed7/pkg/logstructured/logstructured.go#L156
- ensure old object is always filled to fix/workaround `nats/backend.go` and every other future plugins working incorrectly.

mhmxs:
I fixed in my plugin

PR #397, opened 2025-01-10
Introduce max items to list

mhmxs:
Fetching more than 10000 records from the db should make Kubernetes unresponsive. This change prevents Kine users to shoot themself in the feet. At some places Kubernetes also uses `10000` as the maximum number of items to fetch in once.

WDYT?

### DIFF
--- a/pkg/server/list.go
+++ b/pkg/server/list.go
@@ -3,12 +3,15 @@ package server
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/sirupsen/logrus"
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
 )
+
+const maxItems = int64(10000)
 
 func (l *LimitedServer) list(ctx context.Context, r *etcdserverpb.RangeRequest) (*RangeResponse, error) {
 	if len(r.RangeEnd) == 0 {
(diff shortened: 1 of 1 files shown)


pkg/server/list.go
```diff
@@ -57,6 +61,10 @@ func (l *LimitedServer) list(ctx context.Context, r *etcdserverpb.RangeRequest)
 		rev, resp.Count, err = l.backend.Count(ctx, prefix, start, revision)
 		logrus.Tracef("LIST COUNT key=%s, end=%s, revision=%d, currentRev=%d count=%d", r.Key, r.RangeEnd, revision, rev, resp.Count)
 		resp.Header = txnHeader(rev)
+
+		if err == nil && resp.Count > maxItems && (r.Limit == 0 || r.Limit > maxItems) && prefix == start {
+			return nil, errors.New("dataset size exceeds the allowed limit, please use limit or rage end to reduce the number of items")
+		}
```
brandond, line 67: etcd doesn't enforce this limitation as far as I know, so in general I would be against doing something that etcd doesn't. If the apiserver does this for us on the Kubernetes side, why should we do it here?

mhmxs: @brandond because users should forget it, and a wrong query should kill the cluster. I can totally agree with your point, i killed my cluster a few times because i forgot filtering or limiting my query :D

mhmxs:
> etcd doesn't enforce this limitation as far as I know, so in general I would be against doing something that etcd doesn't. 

make sense

PR #400, opened 2025-01-21
Bump k8s dependencies to v0.29.12

dereknola:
## Changes
- Now that 1.28 is EOL, we can move to the latest version of 1.29
- Remove `go toolchain` again. It keeps getting added, even those we don't need it.


brandond (approved):

brandond:
All we're using from kubernetes is `wait.PollInfiniteWithContext`  and `workqueue.NewDelayingQueue`. I wonder if we should work towards removing k8s deps from kine. Wait would be pretty easy to replace, WorkQueue is larger and would be a bit more work.

dereknola:
Good idea, I'll open an issue

vitorsavian (approved):

manuelbuil (approved):

Issue #357, opened 2024-11-08
Kine skips compact intervals if compaction fails to complete


brandond:
If the compaction loop fails for any reason, the rows will be compacted and the compact-rev key will be updated, but the expected compact-rev key stored in memory won't be updated - so it thinks that some other node has compacted, and skips the following interval.

This has been reported several times:
* https://github.com/k3s-io/k3s/discussions/10626#discussioncomment-10231082
* https://github.com/k3s-io/k3s/discussions/11251#discussioncomment-11183233

The first instance was in an odd multi-master Galera cluster, but the second was on plain old sqlite.

This is because if any compaction fails, we restart the outer loop:
https://github.com/k3s-io/kine/blob/c1b2bd81f697c6b7aec85ea2562bcbcdfb981307/pkg/logstructured/sqllog/sql.go#L155-L157
without recording any of the work done by prior successful iterations of the inner loop:
https://github.com/k3s-io/kine/blob/c1b2bd81f697c6b7aec85ea2562bcbcdfb981307/pkg/logstructured/sqllog/sql.go#L165-L167

We should fix that, but we should also figure out how to better handle locking errors when trying to compact.

For sqlite at least, this may be related to go-sqlite3's BeginTX ignoring TxOptions:
* https://github.com/mattn/go-sqlite3/issues/685

This is BAD, as the default behavior of sqlite transactions is to... not actually start a transaction:
https://sqlite.org/forum/info/c3cb9524bef62b67#forum11484
> A bare BEGIN (as in BEGIN DEFERRED) does not start a transaction. It turns off the auto-commit machinery so that the transaction commenced by the next statement is not automatically committed at the end of the execution of that statement. If that statement is a "read" statement, then the transaction is a read transaction. If that statement is a "write" statement, then the transaction is a write transaction. BEGIN IMMEDIATE and BEGIN EXCLUSIVE both turn off the auto-commit machinery and start a transaction (write or exclusive respectively)

brandond:
For sqlite3, I guess all we can do is change the global transaction lock type by setting `_txlock=immediate` in the DSN parameters:
https://github.com/mattn/go-sqlite3/blob/82bc911e85b3def2940e41767480745bcbb6ef45/sqlite3.go#L1162-L1174

PR #414, opened 2025-02-16
fix: only update revisions if currentRev is updated

svanharmelen:
We encountered issues with our setup after updating Kine to version v0.13.7. The issues we encountered where similar to the ones described in #357 and were supposed to be fixed in #360 and #365.

While I noticed that our custom SQLite DSN was missing a few options that seem to help remedy the issues, I still had the impression something else was not playing nice as well.

So after building a local test setup and using the `jobloader` example to create a small load generator that matched our usage profile (guess in hindsight that wasn't even necessary) I was able to reproduce the issue.

We noticed in the test logs that at a certain moment in time the compaction process stopped all together, just as it does in our production setup. Before it stopped it already failed a few times (each time with the `database is locked` error), but each time (except for the last time) it resumed/tried again after 5 minutes.

What caught my eye when comparing logs with the test setup and our production logs is that it only stopped completely when there wasn't a single successful compaction done before the `database is locked` error was received. When looking at the code with that in mind it was easy to see the minor mistake in the code that was responsible for this behavior.

As `currentRev` is declared in the loop, it's reset to 0 on each iteration. So if it's not updated at least once `targetCompactRev` will be set to 0 at the end of the loop and will prevent any compaction to be done from that point on.

### DIFF
--- a/pkg/logstructured/sqllog/sql.go
+++ b/pkg/logstructured/sqllog/sql.go
@@ -173,11 +173,15 @@ func (s *SQLLog) compactor(interval time.Duration) {
 			}
 		}
 
-		// Store the final results for this compact interval.
-		// Note that one or more of the small-batch compact transactions
-		// may have succeeded and moved the compact revision forward, even if err is non-nil.
-		compactRev = compactedRev
-		targetCompactRev = currentRev
+		// Only store the final results for this compact interval if currentRev is
+		// updated to the current compact revision.
+		//
+		// Note that one or more of the small-batch compact transactions may have
+		// succeeded and moved the compact revision forward, even if err is non-nil.
+		if currentRev > 0 {
+			compactRev = compactedRev
+			targetCompactRev = currentRev
+		}
 
 		// ErrCompacted indicates that no further work is necessary - either compactRev changed since the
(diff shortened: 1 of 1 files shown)


brandond:
Thanks for the PR. Can you comment more on what options you were using in your DNS that caused this to occur? Were you able to find any way to reproduce this other than having database locking issues during the very first compaction attempt?

svanharmelen:
We were using `sqlite:///path/to/file?mode=rwc&_journal=WAL&cache=shared`, so we didn't have the `busy_timeout` and `txlock` options.

After reading a few recent issues and PR's we now updated it to `sqlite:///path/to/file?_journal=WAL&cache=shared&_busy_timeout=30000&_txlock=immediate`.

We didn't try to reproduce it using any other errors then the `database is locked` error, but considering the current code the same bug could/should be triggered by any SQL error happening during the first compaction attempt.

At least that is how I read and understand the code 😊

brandond (approved):

svanharmelen:
Thanks for merging this one @brandond 👍🏻 Do you have any (rough) estimate when to expect a new release? We use `k0s` so I need to create a PR there as well to update Kine in their build before we can update to `k8s` 1.32.x...

PR #417, opened 2025-02-18
Remove direct use of deprecated `wait` function

dereknola:
Also has the additional benefit of removing 1 of the K8s direct dependencies.

### DIFF
--- a/pkg/server/watch.go
+++ b/pkg/server/watch.go
@@ -11,7 +11,6 @@ import (
 	"go.etcd.io/etcd/api/v3/etcdserverpb"
 	"go.etcd.io/etcd/api/v3/mvccpb"
 	clientv3 "go.etcd.io/etcd/client/v3"
-	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 var watchID int64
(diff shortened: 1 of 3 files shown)


pkg/server/watch.go
```diff
@@ -39,7 +38,19 @@ func (s *KVServerBridge) Watch(ws etcdserverpb.Watch_WatchServer) error {
...
+
+		for {
+			select {
+			case <-ws.Context().Done():
+				return
+			case <-ticker.C:
+				w.ProgressIfSynced(ctx)
+			}
+		}
+	}(ws.Context())
```
brandond, line 53: ugh I hate how ugly this inline func is compared to the wait helpers. I think we used to have something similar to this and I replaced it with the wait.

dereknola: Well the wait utility function itself is deprecated. Would you rather keep the K8s dependency, but move to the newer wait function, the `PollUntilContextCancel`?

brandond: Maybe a utility function in https://github.com/k3s-io/kine/tree/master/pkg/util? idk I just hate sprinkling goroutine ticker loop functions around otherwise tidy functions.

brandond (approved):

PR #418, opened 2025-02-20
return url encoding hint with url parse error

### LINKED ISSUE #411: error "net/url: invalid userinfo" with some passwords
I'm not really sure if this is actually a bug but definitely a nasty gotcha - raising it here for awareness.

I'm using endpoint `postgres://clusteradmin:xxxxxxxxxxxx@example.com:5432/kine?sslmode=prefer` and seeing error `net/url: invalid userinfo`

My password contains a `>` character. If I remove that, the error goes away.

I think I've hit this issue: https://stackoverflow.com/questions/48671938/go-url-parsestring-fails-with-certain-user-names-or-passwords

> It [Go] now expects the user info to be percent encoded string in order to handle special characters.

ianb-mp:
Resolves #411 

Wrap parse errors with a hint: `error parsing URL. Ensure that any embedded values such as username & password have been URL encoded: <wrapped error>`

### DIFF
--- a/pkg/util/url.go
+++ b/pkg/util/url.go
@@ -8,5 +8,5 @@ import (
 
 func ParseURL(urlStr string) (*url.URL, error) {
 	u, err := url.Parse(urlStr)
-	return u, errors.Wrap(err, "error parsing URL. Ensure that any embedded values such as username & password have been URL encoded")
+	return u, errors.Wrap(err, "failed to parse datastore DSN URL; ensure that parameters containing special characters use percent-encoding")
 }
--- a/pkg/util/url.go
+++ b/pkg/util/url.go
@@ -1,15 +1,12 @@
 package util
 
 import (
-	"fmt"
 	"net/url"
+
+	"github.com/pkg/errors"
 )
 
 func ParseURL(urlStr string) (*url.URL, error) {
 	u, err := url.Parse(urlStr)
-	if err != nil {
-		return nil, fmt.Errorf("error parsing URL. Ensure that any embedded values such as username & password have been URL encoded: %w", err)
-	}
-
-	return u, nil
+	return u, errors.Wrap(err, "error parsing URL. Ensure that any embedded values such as username & password have been URL encoded")
 }
--- a/pkg/util/url.go
+++ b/pkg/util/url.go
@@ -1,0 +1,15 @@
+package util
+
+import (
+	"fmt"
+	"net/url"
+)
+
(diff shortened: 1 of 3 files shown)


pkg/util/url.go
```diff
@@ -0,0 +1,15 @@
...
+	"net/url"
+)
+
+func ParseURL(urlStr string) (*url.URL, error) {
+	u, err := url.Parse(urlStr)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing URL. Ensure that any embedded values such as username & password have been URL encoded: %w", err)
+	}
+
+	return u, nil
```
brandond, line 14: Just use errors.Wrap from `github.com/pkg/errors`. It will return nil if the passed-in error is nil.

```suggestion
	u, err := url.Parse(urlStr)
	return u, errors.Wrap(err, "ensure that DSN parameters such as usernames or passwords containing special characters have been URL encoded")
	}

	return u, nil
```

brandond (changes requested):

pkg/util/url.go
ianb-mp: Done

pkg/util/url.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"net/url"
+
+	"github.com/pkg/errors"
+)
+
+func ParseURL(urlStr string) (*url.URL, error) {
+	u, err := url.Parse(urlStr)
+	return u, errors.Wrap(err, "error parsing URL. Ensure that any embedded values such as username & password have been URL encoded")
```
brandond, line 11: The convention is for errors to not be complete sentences, just fragments - as they are generally joined or appended when passed up the stack. It is also not clear what URL is being referred to here - this is specifically the DSN and we should refer to it as such. 

Also, the RFC that standardizes this calls it "percent-encoding" and we should do likewise: https://datatracker.ietf.org/doc/html/rfc3986#section-2.1

brandond, line 11: how about.
```suggestion
	return u, errors.Wrap(err, "failed to parse datastore DSN URL; ensure that parameters containing special characters use percent-encoding")
```

brandond (changes requested):

pkg/util/url.go
ianb-mp: Agreed.

ianb-mp: lgtm

brandond (approved):

PR #426, opened 2025-03-01
Fix markdown to use equals as header.

theohogberg:
This is standard markdown syntax. Removes the '=====...' in markdown viewers (github etc.)


dereknola (approved):

brandond (approved):
sure.

PR #447, opened 2025-04-02
Split sqlite conformance and load testing

dereknola:
Saves about 5min in CI if we parallelize these two tests.


brandond:
The perf test numbers won't be valid if there's other stuff going on in the background.

We don't really have a good way to compare output from different runs other than just eyeballing it, but the idea behind the perf tests was that we could identify any positive or negative perf impacts of a change by comparing output from PR against the baseline of previous merge CI.

PR #454, opened 2025-04-09
Fix CI and add riscv64 build

### LINKED ISSUE #453: Kine does no longer ship with SQlite
Since `v0.13.11`, new release process no longer builds Kine with CGO, thus only `amd64` ships with SQlite driver.

shanduur:
Fixes issues with static linking.
Fixes issues with missing CGO/SQLite.
Adds two new architectures to release: `riscv64` ~and `ppc64le`~.

Closes #453

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,7 @@ ARG TAG
 ARG DRONE_TAG
 ARG ARCH=amd64
 ENV ARCH=${ARCH}
+ENV CGO_ENABLED=1
 
 COPY ./scripts/build ./scripts/version ./scripts/
 COPY ./go.mod ./go.sum ./main.go ./
@@ -63,6 +64,7 @@ ENV SRC_DIR=/go/src/github.com/k3s-io/kine
 WORKDIR ${SRC_DIR}/
 ARG TARGETOS
 ARG TARGETARCH
+ENV CGO_ENABLED=1
 COPY ./scripts/buildx ./scripts/version ./scripts/
 COPY ./go.mod ./go.sum ./main.go ./
 COPY ./pkg ./pkg
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,7 +53,7 @@ EXPOSE 2379/tcp
 USER nobody
 ENTRYPOINT ["/bin/kine"]
 
-FROM --platform=$BUILDPLATFORM golang:1.24-alpine3.20 AS multi-arch-build
+FROM golang:1.24-alpine3.20 AS multi-arch-build
 RUN apk -U add bash coreutils git gcc musl-dev vim less curl wget ca-certificates
 # go imports version gopls/v0.15.3
 # https://github.com/golang/tools/releases/latest
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ build:
 		-f Dockerfile --target=binary --output=. .
 
 .PHONY: multi-arch-build
-PLATFORMS = linux/amd64,linux/arm64,linux/arm/v7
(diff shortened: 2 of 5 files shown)


scripts/buildx
```diff
@@ -23,11 +23,11 @@ if [ "$TARGETOS" != "linux" ]; then
     OPT_OS="-${TARGETOS}"
 fi
 
-if [ "$TARGETOS" = "Linux" ]; then
-    OTHER_LINKFLAGS="-extldflags -static -s"
+if [ "$TARGETOS" = "linux" ]; then
```
brandond, line 26: this needs to be fixed in `scripts/build` as well

brandond (changes requested):
thanks for the fix! One nit on the TARGETOS fix.

scripts/buildx
shanduur: Fixed!

brandond (approved):

Dockerfile
```diff
@@ -53,7 +54,7 @@ EXPOSE 2379/tcp
 USER nobody
 ENTRYPOINT ["/bin/kine"]
 
-FROM --platform=$BUILDPLATFORM golang:1.24-alpine3.20 AS multi-arch-build
```
dereknola, line 56: This needs to remain, its what allows the cross-compiling only the gobinary, instead of using QEMU to emulate the whole image build, which is incredibly slow in GHA.

Makefile
```diff
@@ -24,7 +24,7 @@ build:
 		-f Dockerfile --target=binary --output=. .
 
 .PHONY: multi-arch-build
-PLATFORMS = linux/amd64,linux/arm64,linux/arm/v7
```
dereknola, line 27: I don't see any interest in building ppc64le.

dereknola (changes requested):

Dockerfile
shanduur: This means you can't use CGO, and need to drop SQLite or migrate to pure go driver.

Makefile
shanduur: Dropped.

Dockerfile
shanduur: Pure Go implementations will be probably [much slower](https://github.com/cvilsmeier/go-sqlite-bench).

dereknola: When I wrote https://github.com/k3s-io/kine/commit/fde41b68fbd6a5e68a07c87bfda1c7059af74e2b, I didn't see a `CGO_ENABLED` and so assumed we were not using it. I would not have written the pipeline this way if I had known.

This changes everything, the GHA needs to be overhauled rewritten to use either tonistiigi/xx or a build matrix of native amd64/arm64 runners.

shanduur: I'll try to rewrite it using CGO and XX

dereknola: Specifically, emulate arm on arm64 (its near native performance). Riscv64 it was faster to emulate on amd64 than arm64, its just so different from either platform that the more powerful amd64 cores helped.

dereknola: But try and rewrite with XX first, thats the easier solution. If that doesn't work out, I can implement the matrix build in another PR.

dereknola:
I got a branch that should build everything successfully with CGO. See https://github.com/k3s-io/kine/commit/5a89c4cb47c45e06786dbdfbc28e33124dbe3fd2

brandond:
You wanna go with that instead @dereknola ?

dereknola:
I don't want to steal @shanduur thunder, but if they are happy with it. I'll add them as a coauthor since they found this issue.

shanduur:
Yup, go ahead and merge @dereknola !

Issue #452, opened 2025-04-07
Feature Request: Support Custom Table Names Configuration


happy-game:
I came across the closed issue (#402 ) where a user requested the ability to **specify custom table names** due to permissions constraints and business-specific naming requirements. In my own use case, I also require this functionality for similar reasons.

I would like to know **if there are current plans to support this feature**. If so, I would like to contribute a PR to add support for configurable table names (e.g., via environment variables or CLI flags).

brandond:
PR welcome.

phuslu:
I made a suggestion here, https://github.com/k3s-io/kine/issues/141, but in fact, it is very difficult to implement this feature without significant modifications, because `generic`/`drivers` uses a lot of global string concatenation to assemble SQL, so it is too late when we try to parse the table name with parameters.

I made a minimal PoC [here ](https://github.com/phuslu/kine/commits/master/)，This should be the simplest change, but it's a bit hacky.

PR #455, opened 2025-04-10
Support Custom Table Names Configuration

happy-game:
- Add a CLI parameter `--table-name`, with the default value being `kine`.
- Add `TableName` to the `config` in `pkg/endpoint`.
- Add `TableName` to `pkg/drivers/config.go`.
- Support custom table names for `sqlite`, `mysql`, and `pgsql`. 

I have already run `scripts/test` and passed all the tests.
image

#452

### DIFF
--- a/pkg/drivers/generic/generic.go
+++ b/pkg/drivers/generic/generic.go
@@ -30,12 +30,11 @@ const (
 var _ server.Dialect = (*Generic)(nil)
 
 var (
-	columns         = "kv.id AS theid, kv.name AS thename, kv.created, kv.deleted, kv.create_revision, kv.prev_revision, kv.lease, kv.value, kv.old_value"
-	revSQL          string
-	compactRevSQL   string
-	listSQL         string
-	tableName       string
-	quotedTableName string
+	columns       = `kv.id AS theid, kv.name AS thename, kv.created, kv.deleted, kv.create_revision, kv.prev_revision, kv.lease, kv.value, kv.old_value`
+	revSQL        string
+	compactRevSQL string
+	listSQL       string
+	tableName     string
 )
 
 type ErrRetry func(error) bool
@@ -96,8 +95,8 @@ func q(sql, param string, numbered bool) string {
 func (d *Generic) Migrate(ctx context.Context) {
 	var (
 		count     = 0
-		countKV   = d.queryRow(ctx, "SELECT COUNT(*) FROM key_value")
-		countKine = d.queryRow(ctx, "SELECT COUNT(*) FROM "+quotedTableName)
+		countKV   = d.queryRow(ctx, `SELECT COUNT(*) FROM key_value`)
+		countKine = d.queryRow(ctx, `SELECT COUNT(*) FROM "`+tableName+`"`)
 	)
 
 	if err := countKV.Scan(&count); err != nil || count == 0 {
@@ -110,7 +109,7 @@ func (d *Generic) Migrate(ctx context.Context) {
 
 	logrus.Infof("Migrating content from old table")
 	_, err := d.execute(ctx,
-		`INSERT INTO `+quotedTableName+`(deleted, create_revision, prev_revision, name, value, created, lease)
+		`INSERT INTO "`+tableName+`"(deleted, create_revision, prev_revision, name, value, created, lease)
 					SELECT 0, 0, 0, kv.name, kv.value, 1, CASE WHEN kv.ttl > 0 THEN 15 ELSE 0 END
 					FROM key_value kv
 						WHERE kv.id IN (SELECT MAX(kvd.id) FROM key_value kvd GROUP BY kvd.name)`)
@@ -151,21 +150,21 @@ func validateTableName(customTableName string) error {
 func buildSQLStatements() (rev, compactRev, list string) {
 	rev = fmt.Sprintf(`
 		SELECT MAX(rkv.id) AS id
-		FROM %s AS rkv`, quotedTableName)
+		FROM "%s" AS rkv`, tableName)
 
 	compactRev = fmt.Sprintf(`
 		SELECT MAX(crkv.prev_revision) AS prev_revision
-		FROM %s AS crkv
-		WHERE crkv.name = 'compact_rev_key'`, quotedTableName)
+		FROM "%s" AS crkv
+		WHERE crkv.name = 'compact_rev_key'`, tableName)
 
 	list = fmt.Sprintf(`
 		SELECT *
 		FROM (
 			SELECT (%s), (%s), %s
-			FROM %s AS kv
+			FROM "%s" AS kv
 			JOIN (
 				SELECT MAX(mkv.id) AS id
-				FROM %s AS mkv
+				FROM "%s" AS mkv
 				WHERE
 					mkv.name LIKE ?
 					%%s
@@ -176,7 +175,7 @@ func buildSQLStatements() (rev, compactRev, list string) {
 				?
 		) AS lkv
 		ORDER BY lkv.thename ASC
-		`, rev, compactRev, columns, quotedTableName, quotedTableName)
+		`, rev, compactRev, columns, tableName, tableName)
 
 	return rev, compactRev, list
 }
@@ -208,16 +207,6 @@ func Open(ctx context.Context, driverName, dataSourceName string, connPoolConfig
 	}
 
 	tableName = customTableName
-
-	// In case of MySQL, we need to quote the table name using ` `
-	// In case of SQLite and Postgres, we need to quote the table name using " "
-	switch driverName {
-	case "mysql":
-		quotedTableName = "`" + tableName + "`"
-	default:
-		quotedTableName = `"` + tableName + `"`
-	}
-
 	revSQL, compactRevSQL, listSQL = buildSQLStatements()
 
 	for i := 0; i < 300; i++ {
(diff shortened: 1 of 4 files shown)


pkg/drivers/generic/generic.go
```diff
@@ -197,57 +213,64 @@ func Open(ctx context.Context, driverName, dataSourceName string, connPoolConfig
...
+	d := &Generic{
+		DB:        db,
+		TableName: tableName,
+	}
+
+	revSQL := d.RevSQL()
+	compactRevSQL := d.CompactRevSQL()
+	listSQL := d.ListSQL()
+
+	d.GetRevisionSQL = q(fmt.Sprintf(`
```
brandond, line 225: Why did you switch to setting fields on the var instead of setting them in the variable definition? Can you revert this? It makes the change much harder to review.

pkg/drivers/sqlite/sqlite.go
```diff
@@ -33,15 +33,15 @@ var (
...
-		`CREATE INDEX IF NOT EXISTS kine_name_id_index ON kine (name,id)`,
-		`CREATE INDEX IF NOT EXISTS kine_id_deleted_index ON kine (id,deleted)`,
-		`CREATE INDEX IF NOT EXISTS kine_prev_revision_index ON kine (prev_revision)`,
-		`CREATE UNIQUE INDEX IF NOT EXISTS kine_name_prev_revision_uindex ON kine (name, prev_revision)`,
+			)`, tableName),
+		fmt.Sprintf(`CREATE INDEX IF NOT EXISTS %s_name_index ON %s (name)`, tableName, tableName),
+		fmt.Sprintf(`CREATE INDEX IF NOT EXISTS %s_name_id_index ON %s (name,id)`, tableName, tableName),
+		fmt.Sprintf(`CREATE INDEX IF NOT EXISTS %s_id_deleted_index ON %s (id,deleted)`, tableName, tableName),
+		fmt.Sprintf(`CREATE INDEX IF NOT EXISTS %s_prev_revision_index ON %s (prev_revision)`, tableName, tableName),
+		fmt.Sprintf(`CREATE UNIQUE INDEX IF NOT EXISTS %s_name_prev_revision_uindex ON %s (name, prev_revision)`, tableName, tableName),
```
brandond, line 41: There's a ton of new use of `fmt.Sprintf`. This isn't in the hot path, but some of it is. Could we instead just do this:

Note that I am also concerned about special characters in table names. We probably need to do some cleaning/validation on the table name if we are just going to blindly use it in strings.

```suggestion
		`CREATE INDEX IF NOT EXISTS `+tableName+`_name_index ON `+tableName+` (name)`,
		`CREATE INDEX IF NOT EXISTS `+tableName+`_name_id_index ON `+tableName+` (name,id)`,
		`CREATE INDEX IF NOT EXISTS `+tableName+`_id_deleted_index ON `+tableName+` (id,deleted)`,
		`CREATE INDEX IF NOT EXISTS `+tableName+`_prev_revision_index ON `+tableName+` (prev_revision)`,
		`CREATE UNIQUE INDEX IF NOT EXISTS `+tableName+`_name_prev_revision_uindex ON `+tableName+` (name, prev_revision)`,
```

brandond (changes requested):

pkg/drivers/generic/generic.go
happy-game: Apologies for the confusion. I structured it this way because `revSQL`, `compactRevSQL`, and `listSQL` are dynamically generated by calling `RevSQL()` (and related functions), which internally depend on `d.TableName`. Since the table name is required to construct these SQL statements, I needed to set `d.TableName` first before defining the variables.  

Alternatively, I could move `TableName` out of the `Generic` struct and treat it as a global variable(like `columns`) if you think that would be clearer. Would you prefer that approach?

Happy to revise this based on your suggestion.

happy-game: Done. I have reverted this. I set `revSQL`, `compactRevSQL`, and `listSQL` as package-level variables and assigned values to them according to `tableName` in the `Open` function.

pkg/drivers/sqlite/sqlite.go
happy-game: Done.
1. I used the `+` operator to concatenate strings instead of `fmt.Sprintf`, but in the `generic.Open` function, `fmt.Sprintf` is still used because that part originally used `fmt.Sprintf`, so as to maintain consistency.
2. Imposed restrictions on `tableName`:
    - Restrict the table name to be composed only of letters, numbers, and underscores.
    - Restrict the length of the table name to be less than 32 (in `mysql`, the length of `table` and `index` is restricted to be less than 64) [MySQL doc](https://dev.mysql.com/doc/refman/8.4/en/identifier-length.html).
    - Restrict that it can only start with a letter (the restriction in `mariadb`) [MariaDB doc](https://mariadb.com/kb/en/columnstore-naming-conventions/) .
3. Enclose `tableName` with quotes to prevent the table name from being a reserved word like `SELECT`.

happy-game:
I'm a bit confused because I noticed that there was an issue with the CI when it was loading the configmap for SQLite, and the error occurred during the inter-container communication phase. However, it worked normally for MySQL and PostgreSQL.
```
[sqlite] Waiting for kubeconfig to become available...
[sqlite] Running configmap load generation script
[sqlite] [234]	I Connected to https://172.17.0.22:6443 - v1.29.4+k3s1
[sqlite] [232]	I Connected to https://172.17.0.22:6443 - v1.29.4+k3s1
[sqlite] [231]	I Connected to https://172.17.0.22:6443 - v1.29.4+k3s1
[sqlite] [233]	I Connected to https://172.17.0.22:6443 - v1.29.4+k3s1
[sqlite] [233]	W Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'RemoteDisconnected('Remote end closed connection without response')': /api/v1/namespaces/load-test/configmaps/test-45
[sqlite] [232]	W Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'RemoteDisconnected('Remote end closed connection without response')': /api/v1/namespaces/load-test/configmaps/test-55
[sqlite] [231]	W Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'RemoteDisconnected('Remote end closed connection without response')': /api/v1/namespaces/load-test/configmaps/test-73
[sqlite] [234]	W Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'RemoteDisconnected('Remote end closed connection without response')': /api/v1/namespaces/load-test/configmaps/test-05
```
After the CI failed, I retested both locally and in GitHub Codespaces. In both cases, the configmap could be loaded normally for performance testing, and all subsequent tests passed.

pkg/drivers/pgsql/pgsql.go
```diff
@@ -28,9 +28,12 @@ const (
...
 
-var (
-	schema = []string{
-		`CREATE TABLE IF NOT EXISTS kine
+var createDB = `CREATE DATABASE "%s";`
+
+func getSchema(tableName string) []string {
+	quotedTableName := `"` + tableName + `"`
+	return []string{
+		`CREATE TABLE IF NOT EXISTS ` + quotedTableName + `
```
brandond, line 36: This isn't quite what I had in mind...

Lets just be consistent about how we build strings for the table name - just always do:
```golang
`CREATE TABLE IF NOT EXISTS "` + tableName + `"
```

But then we restrict the table name to `[0-9a-zA-Z$_]`. Someone may complain about not being able to characters from some other locale but it's easier than trying to ensure we properly quote all the possible values. ref: https://github.com/golang/go/issues/18478

pkg/app/app.go
```diff
@@ -111,6 +111,12 @@ func New() *cli.App {
 			Destination: &config.EmulatedETCDVersion,
 			Value:       "3.5.13",
 		},
+		&cli.StringFlag{
+			Name:        "table-name",
+			Usage:       "The table name for the selected backend. Defaults to 'kine'.",
+			Destination: &config.TableName,
+			Value:       "kine",
+		},
```
brandond, line 119: Put this up next to the `endpoint` flag, so that they appear together in help.

brandond (changes requested):
couple more nits

pkg/drivers/pgsql/pgsql.go
happy-game: I have also considered this approach. The use of `quotedTableName` is a trade - off I made , due to the following reasons. :
1. In MySQL, [The identifier quote character is the backtick](https://dev.mysql.com/doc/refman/8.4/en/identifiers.html). In this case, we need to modify the quotation marks used to define strings, which will make multi-line SQL statements less intuitive.
2. In generic/generic.go, we need to **consider multiple SQL drivers**, and we have to use more `if` statements to decide which kind of quotation marks to use, which will increase redundancy.

In order to make MySQL also support using double quotes (" ") to quote identifiers, I need to execute 
```sql
SET SESSION sql_mode = 'ANSI_QUOTES';
``` 
at the start of the session.

This way, we can uniformly use double quotes ("") and no longer need to use the `quotedTableName`. 

Do you agree with this?

happy-game:
Apologies, I amended a small change (adjusting CLI parameter position) onto a previous commit. This changed the subsequent commit hashes. I'm very sorry that this increased your review workload.

happy-game:
Hi @brandond ,I've fixed the issue you mentioned previously. Could you please review the code at your convenience?
Sorry for the extra work caused by my previous mistake.

brandond:
This looks good but I'd like a chance to look it over a bit more closely to think if there's some simpler way to do it without adding so much string mangling and new functions. I'll try to get to that next week.

happy-game:
I have rebased onto the latest master branch.

brandond:
1. This needs rebasing, sorry
2. I know our SQL generation is already a mess, but honestly I think we need to refactor it a bit more before we do anything like this - it makes what is already a confusing mess of SQL snippets even more obtuse. Can you think of a good way to do this that's perhaps a bit cleaner?

happy-game:
If you still feel this feature is necessary, I'm happy to implement it. However, I would prefer to start fresh on the latest commit and open a new pull request.
On the one hand, I haven't looked at this PR in a while, and it's a bit of a mess. On the other hand, writing the code from scratch would likely be easier than trying to rebase and resolve all the conflicts.
I'll give it more thought this time to ensure the new PR is much cleaner and more focused.

PR #464, opened 2025-04-17
Configurable compaction and observability enhancements

yegortokmakov:
Hi! I'm pushing upstream some of our changes. All of the changes have been tested in our clusters and are backwards compatible.

1. The compact operation configuration is now exposed via CLI flags: `--compact-timeout`, `--compact-min-retain`, `--compact-batch-size`, and `--poll-batch-size`. We run quite a few clusters and wanted an easier way to tune these settings to optimize our database load. Default values and behavior haven't changed.

2. Added the flag `--metrics-ignore-tls-config`, which, if set to `true`, will expose the `/metrics` endpoint over HTTP without TLS, regardless of the main endpoint's configuration. Not sure if the naming of the flag is perfect, but it simply prevents copying the TLS config in this line: `metricsConfig.ServerTLSConfig = config.ServerTLSConfig`.

3. Added the `--log-format` flag with possible values: `plain` (default) and `json`. Plain logging is the current formatting configuration in Kine.

4. Added logging for "duplicate key value violates unique constraint" errors. Retriable errors are logged as warnings, and non-retriable ones are logged as errors.

5. Added the metric `kine_insert_errors_total` with labels `key` and `retriable` (`"true"`/`"false"`).

6. Adjusted the behavior of slow SQL reporting: queries longer than 5 seconds are now logged as warnings instead of info.

6. Added a new flag `--compact-interval-jitter` to apply a jitter to the interval in order to distribute the load more evenly when multiple Kine instances are set up with the same interval.

I understand it's easier to review smaller PRs, so let me know if you'd prefer this split into smaller pull requests.

### DIFF
--- a/pkg/drivers/generic/generic.go
+++ b/pkg/drivers/generic/generic.go
@@ -444,14 +444,14 @@ func (d *Generic) Insert(ctx context.Context, key string, create, delete bool, c
 		err = row.Scan(&id)
 
 		if err != nil && d.InsertRetry != nil && d.InsertRetry(err) {
-			logrus.Warnf("unique violation retry for key %v: %v", key, err)
+			logrus.Warnf("retriable insert error: duplicate key %v violates unique constraint: %v", key, err)
 			metrics.InsertRetriesTotal.WithLabelValues().Inc()
 			wait(i)
 			continue
 		}
 
 		if err != nil {
-			logrus.WithField("key", key).WithField("createRevision", createRevision).WithField("previousRevision", previousRevision).Errorf("unique violation error for key %v: %v", key, err)
+			logrus.WithField("key", key).WithField("createRevision", createRevision).WithField("previousRevision", previousRevision).Errorf("insert error: duplicate key %v violates unique constraint: %v", key, err)
 		}
 
 		return id, err
--- a/pkg/drivers/generic/generic.go
+++ b/pkg/drivers/generic/generic.go
@@ -445,13 +445,13 @@ func (d *Generic) Insert(ctx context.Context, key string, create, delete bool, c
 
 		if err != nil && d.InsertRetry != nil && d.InsertRetry(err) {
 			logrus.Warnf("retriable insert error for key %v: %v", key, err)
-			metrics.InsertErrorsTotal.WithLabelValues(key, "true").Inc()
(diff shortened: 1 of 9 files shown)


pkg/drivers/generic/generic.go
```diff
@@ -442,10 +442,19 @@ func (d *Generic) Insert(ctx context.Context, key string, create, delete bool, c
 	for i := uint(0); i < 20; i++ {
 		row := d.queryRow(ctx, d.InsertSQL, key, cVal, dVal, createRevision, previousRevision, ttl, value, prevValue)
 		err = row.Scan(&id)
+
 		if err != nil && d.InsertRetry != nil && d.InsertRetry(err) {
+			logrus.Warnf("retriable insert error for key %v: %v", key, err)
+			metrics.InsertErrorsTotal.WithLabelValues(key, "true").Inc()
```
brandond, line 448: I would recommend against including the key in this metric. Log it if you want, but including it in metrics can lead to excessively large metric cardinality. Ref: https://prometheus.io/docs/practices/naming/#labels

> Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

pkg/metrics/metrics.go
```diff
@@ -30,23 +30,35 @@ var (
 		Name: "kine_compact_total",
 		Help: "Total number of compactions",
 	}, []string{"result"})
+
+	InsertErrorsTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "kine_insert_errors_total",
+		Help: "Total number of insert retries due to unique constraint violations",
+	}, []string{"key", "retriable"})
```
brandond, line 37: same nit re including the key in the labels

brandond (changes requested):
thanks for the PR! One nit on the new metric, the other things all LGTM.

brandond (approved):

cagataygurturk:
Very useful improvement @yegortokmakov . When can we expect it to be merged and released?

dereknola (approved):

yegortokmakov:
hi @dereknola @brandond , thanks for the review! anything on my side I should do to get this released? thanks!

yegortokmakov:
hi @brandond , thanks for merging! how often do you cut releases? anything I can do to help?

PR #481, opened 2025-05-20
Skip logging of ErrKeyExists during insert

mhmxs:
This change eliminates logging of `duplicate key value violates unique constraint "kine_name_prev_revision_uindex" (SQLSTATE 23505)` (in case of postgresql). There are two reasons to do this:

- ErrKeyExists is a special error, it is not logged when happens on mysql for example.
- ErrKeyExists suspects client error, and handled by client, so it makes sense to skip at Kine side.

### DIFF
--- a/pkg/drivers/generic/generic.go
+++ b/pkg/drivers/generic/generic.go
@@ -450,7 +450,7 @@ func (d *Generic) Insert(ctx context.Context, key string, create, delete bool, c
 			continue
 		}
 
-		if err != nil {
+		if err != nil && (d.TranslateErr == nil || d.TranslateErr(err) != server.ErrKeyExists) {
 			metrics.InsertErrorsTotal.WithLabelValues("false").Inc()
 			logrus.WithField("key", key).WithField("createRevision", createRevision).WithField("previousRevision", previousRevision).Errorf("insert error for key %v: %v", key, err)
 		}


brandond:
@yegortokmakov I believe you JUST started logging this in https://github.com/k3s-io/kine/pull/464 - can you provide some commentary on why you found this useful?

mhmxs:
> they are definitely a symptom of something wrong in the setup

For example, Kubernetes E2E tests generates lots of messages. The tests are handling those failures, so they don't have any purpose, except it is harder to read the logs.

brandond:
If you're only seeing errors on e2e tests that specifically do bad things, then I'm less likely to merge a PR that just churns the logging back off again.

mhmxs:
> If you're only seeing errors on e2e tests that specifically do bad things, then I'm less likely to merge a PR that just churns the logging back off again.

@brandond I'm a developer, so the only workload i run are tests. I don't think the E2E test does anything wrong. It just handle already exists error. What i can say, earlier it didn't log this messages, and now it is annoying to read them, because they are client side issues and not Kine errors :)

yegortokmakov:
> > they are definitely a symptom of something wrong in the setup
> 
> 
> 
> For example, Kubernetes E2E tests generates lots of messages. The tests are handling those failures, so they don't have any purpose, except it is harder to read the logs.

What's your setup for the tests? Do you have an example I can try to reproduce? I'm curious what can cause it. 

In the end, it might be that the Postgres instance used in the test suite doesn't have enough resources to keep up.

mhmxs:
> The error simply doesn't happen on other databases

I see, but the error i have seen was related to `kine_name_prev_revision_uindex` and not PK collision. That unique constraint is existing on mysql too. The log message is wrapped by `err != nil`, so it logs every error. But if you check `dialect.TranslateErr` of both mysql and pgsql, it translates uniqu constraint errors to `server.ErrKeyExists`, and these errors are not logged in Kine. But this log message doesn't take care on translating of errors, so it logs every error, but only on pgsql.

PR #513, opened 2025-09-03
Update kine for compatibility with kubernetes v1.34

brandond:
* Allow apiserver to watch the compact rev key, by redirecting list/get/watch to the substitute key
  The apiserver now watches the compact rev key and prunes caches based on the compact revision.
  Ref: https://github.com/kubernetes/kubernetes/pull/132876
  This is a follow-up to: **Add support for apiserver-managed compact** https://github.com/k3s-io/kine/commit/6c30f90141376d669405611a29220cde98f35717
  In order for the 1.34+ apiserver to work properly, kine should be used with `--compact-interval=0` to let the apiserver manage compaction.
* Add support for `clientv3.WithKeysOnly()` list/get option.
  Ref: https://github.com/kubernetes/kubernetes/pull/132355
  Kubernetes no longer uses `WithCountOnly()` to count keys under a prefix; it now uses `WithKeysOnly()` to list all keys under a prefix and then counts those. On the backend, we now only include the value and prev_value columns as necessary, which should make a number of queries more efficient.

### DIFF
--- a/pkg/drivers/generic/generic.go
+++ b/pkg/drivers/generic/generic.go
@@ -29,8 +29,10 @@ const (
 var _ server.Dialect = (*Generic)(nil)
 
 var (
-	columns = "kv.id AS theid, kv.name AS thename, kv.created, kv.deleted, kv.create_revision, kv.prev_revision, kv.lease, kv.value, kv.old_value"
-	revSQL  = `
+	columns    = "kv.id AS theid, kv.name AS thename, kv.created, kv.deleted, kv.create_revision, kv.prev_revision, kv.lease"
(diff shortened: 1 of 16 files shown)


dereknola (approved):

cwayne18 (approved):

brandond:
Nothing changed here would increase delete the number of delete or insert queries. Insert comes from creation of new revisions - ie create, update, or delete of a Kubernetes resource. Delete comes ONLY from old resource versions being compacted. This is foundational to how Kine works.

If you are seeing more inserts or deletes, then something has changed in how your workloads are interacting with the apiserver.

PR #525, opened 2025-09-27
Return actual updated/deleted count

melgenek:
First of all, thank you for a great project!

I have recently implemented a [Kine for FoundationDB](https://github.com/melgenek/f8n). The implementation uses the Kine grpc server as is and implements an additional [Backend](https://github.com/k3s-io/kine/blob/9262f52d46e4f30d8a8425907dfdf6ebe9656ceb/pkg/server/types.go#L21-L33).

In order to make sure that the implementation works as expected, I've adjusted the [Etcd robustness tests](https://github.com/etcd-io/etcd/tree/main/tests/robustness) for [Kine+FoundationDB](https://github.com/melgenek/f8n/tree/main/tests/api). These tests send a bunch of requests that mimic K8S traffic and check linearizability/serializability/watch guarantees.

Some of the responses that kine sends were not fully compatible, so I had to partially fork Kine.
Namely:
1. delete response for non-existing keys returns `Count=1` even though the value should be `0`
2. The `Deleted` field with a count is not returned for successful deletes.
3. updates always return `Count=1` even if the key doesn't exist at all

It would be amazing if you could accept a small change to the grpc layer to make it compatible with Etcd for the Etcd robustness tests.


melgenek:
>  I'm really not sure why the response has an explicit length field when the length can be calculated from the array size, as Kubernetes does...

The Etcd robustness test has workloads for both generic etcd, and specifically kubernetes. 
The framework for evaluating results is the same, but the requests are different. For kubernetes there are only requests sent that Etcd folks extracted from tracing the Api server.

However, in the regular etcd robustness tests, there might be any requests set that etcd supports.
For example, for the delete operation, there are at least two ways to perform a delete:
1. Txn request, that API server sends
2. or a `DeleteRangeRequest` that API server does not use, but other Etcd clients can
`DeleteRangeRequest` has a flag `PrevKv bool`. If `PrevKv=false`, the response would have no kvs, but have a counter.

melgenek:
Makes sense. Thank you!

PR #549, opened 2025-11-07
NATS Driver Conformance Updates

samuelattwood:
- Brings NATS Backend up-to-date
- Resolves issues with Kubernetes Conformance test suite
- Adds read-after-write checks
- Various fixes and enhancements


DeanKamali:
Any update on this? @samuelattwood  

I have some time to dedicate for this, let me know if I can do any reviews or pick up any blockers.

samuelattwood:
There is a small fix I had to make after enabling the apiserver tests that wasn't caught by the Sonobuoy conformance suite.

I have been at KubeCon this week but I will have that up tomorrow or Monday, in addition to some improvements regarding tombstones.

pkg/drivers/nats/new.go
```diff
@@ -93,6 +94,16 @@ func newBackend(ctx context.Context, wg *sync.WaitGroup, connection string, tlsI
...
 		go ns.Start()
+
+		// Shutdown on interrupt.
+		sigch := make(chan os.Signal, 1)
+		signal.Notify(sigch, os.Interrupt)
+		go func() {
+			<-sigch
+			ns.Shutdown()
+			logrus.Infof("embedded NATS server shutdown")
+		}()
```
brandond: No, don't do your own signal handling. This is handled by the CLI or calling application:
* https://github.com/k3s-io/kine/blob/1d571b01fd1e5e08b0276f7d06c3244972acb773/pkg/app/app.go#L225
* https://github.com/k3s-io/kine/blob/1d571b01fd1e5e08b0276f7d06c3244972acb773/pkg/signals/signal.go#L34-L37

You should be using the context and waitGroup that you are provided - call wg.Add(1) when starting, do what you're going to do until the context is done, and then call wg.Done() when you've shut down.

brandond:
1. Can you rebase onto master? CI has been moved from Drone to GHA.
2. Would nats benefit from using the waitgroup to wait until it's shut down properly, or is it OK to just exit as soon as the context is cancelled?

samuelattwood:
@brandond Yep, I will have a rebase up shortly. I'm fixing something with the embedded test I ran into after rebasing.

Waitgroup is probably a good idea, I will add that as well.

brandond:
Sorry, I pushed another round of changes to CI; would you mind rebasing again? I think we're very close to done with this.

brandond:
LGTM! Would you like to do some cleanup on the commit history and fixup/squash the various fix/test/wip commits, or should I just squash when merging?

brandond (approved):

PR #563, opened 2025-12-08
Add optimized SQL for fetching single key in ListCurrent method

### LINKED ISSUE #562: [Bug] Underbar is treated as character wildcard with Get
https://github.com/k3s-io/kine/blob/770d87a8d3380fbd4b9d2c4772e73021df4d01a9/pkg/drivers/generic/generic.go#L52

It's due to the fact that Get uses like to match paths, so maybe we should make list and get use different sql?

wzshiming:
Fixed #562


wzshiming:
Thank you!

PR #557, opened 2025-12-02
Introducing the pure go driver for sqlite

viscropst:
/close #556  
and related to k0sproject/k0s#6730.

PR #573, opened 2025-12-19
Add support for client cert auth on the etcd endpoints

sfackler:
This adds support for basic client cert auth using the same `--trusted-ca-file` flag that ETCD itself uses. I decided to leave the metrics endpoint unauthenticated, but don't have super strong feelings there.

cc #272


brandond:
Please `--sign-off` your commits for DCO! You might as well squash/fixup while you're at it.

brandond:
thanks! I'll probably tag a new release once we get https://github.com/k3s-io/kine/pull/557 merged.

PR #583, opened 2026-01-07
feat: add grpc-health-probe

shanduur:
When running Kine in K8s with TLS, there is no option to configure healthchecks
with TLS. There is a long lasting discussion in K8s enhancments repo, as well
as in kubernetes repo, but the feature has not been implemented.

* https://github.com/kubernetes/enhancements/issues/4939
* https://github.com/kubernetes/kubernetes/issues/119093

Adding grpc-health-probe increases the image size slightly, but allows for
improved health checking in case of TLS enabled deployments.

Signed-off-by: Mateusz Urbanek <mateusz.urbanek.98@gmail.com>

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -202,6 +202,20 @@ func New() *cli.App {
 			Name:    "debug",
 			EnvVars: []string{"KINE_DEBUG"},
 		},
+		&cli.DurationFlag{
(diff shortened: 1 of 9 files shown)


shanduur:
Yeah, that was a quick and dirty example, I've not checked the SQLite params. I've added minimal healthchecks, but I am not sure if the NATS one is correct.

PR #588, opened 2026-01-16
Litestream compatibility options for SQLite backend

### LINKED ISSUE #586: Allow disabling SQLite WAL checkpointing
To use Kine with [Litestream](https://litestream.io) it is required to have WAL checkpointing disabled as Litestream will take care of that. It wouldn't matter so much as Litestream will keep a lock on it that prevents passive checkpoints but because Kine will do a full checkpoint after compacting it will effectively deadlock itself every time for 30 seconds. This also technically does a synchronous write block at compaction which isn't necessarily ideal.

Additionally I found no context why forced checkpointing was added in the first place. SQLite will [automatically do passive checkpointing](https://sqlite.org/wal.html#automatic_checkpoint) at runtime so there shouldn't need to be a reason to do it manually. I found no context in [the commit message](https://github.com/k3s-io/kine/commit/c1edece7772b4ff4f1f03f01d5218fbfe854c0cc) for this change.

There's also an additional TRUNCATE checkpoint after creating the schema which is also quite unnecessary as the WAL file is part of the database.

To fix this use case there are effectively four options.

1. Add a configuration option to disable all checkpointing to use with Litestream
2. Add a configuration option to override the post compaction SQL (so it can be effectively disabled)
3. Change all manual checkpoints to passive but they'll likely be redundant then
4. Revert the commit unless there's a sound reason for it to exist

I believe using Litestream with Kine and K3s is an interesting and useful use case to have instantly recoverable control plane in case of disastrous data loss so hopefully this can be resolved somehow.

hifi:
Fixes #586

Parsing the DS seems like a lot of work just to check a single flag that is fairly unique on its own so didn't do that.

### DIFF
--- a/pkg/drivers/sqlite/sqlite.go
+++ b/pkg/drivers/sqlite/sqlite.go
@@ -7,6 +7,7 @@ import (
 	"database/sql"
 	"fmt"
 	"os"
+	"strings"
 	"sync"
 
 	"github.com/k3s-io/kine/pkg/drivers"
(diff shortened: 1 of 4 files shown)


pkg/drivers/sqlite/sqlite.go
```diff
@@ -56,6 +56,7 @@ func NewVariant(ctx context.Context, wg *sync.WaitGroup, driverName string, cfg
 		}
 		dataSourceName = "./db/state.db?_journal=WAL&cache=shared&_busy_timeout=30000&_txlock=immediate"
 	}
+	noCheckpointing := strings.Contains(dataSourceName, "_kine_no_checkpointing")
```
brandond:
```suggestion
	disableCheckpoint := strings.Contains(dataSourceName, "_kine_disable_wal_checkpoint")
```

hifi:
Since the Litestream specific changes are quite small I'm adding them here. I still need to test that this works but if it does I'd be happy with a compromise that everything else can be configured with the DS except persist WAL that requires the actual SQLite3 connection.

If even more flexibility is needed I could name the SQL driver into `sqlite3-persist-wal` and add it as another driver to Kine but I suspect it's such an exceptional requirement for Litestream that bundling it with it is probably fine.

Let me know if you want me to split these into sparate PRs or if this would be okay. Thanks!

brandond (changes requested):
Thanks! One nit on the additional init function in sqlite.go.

A test for this driver would be appreciated.

hifi:
Indeed, I didn't even notice the driver registration below was already an `init()` function. Moved the sqlite3 driver registration now there as well.

I can't run the test as-is locally and I wasn't sure if you wanted a more comprehensive test where the WAL options are checked and maybe Litestream running or just to run the existing test suite with the driver so I only added that for now. As far as I understand the test suite should complete unless it uses so much disk without WAL checkpointing that it would cause issues.

If checkpointing is needed during tests then it's a bit more involved to add Litestream.

brandond (approved):

PR #653, opened 2026-04-30
Memory Driver

luthermonson:
Hi @brandond long time no see 👋 ... something lived rent free in my mind for the last 7yrs which was an in-memory driver for kine for ephemeral work like e2e tests and ci runs etc. You could kind of do this with sqlite and nats through configuration, but you were just making a translation layer and extra services on top of the memory when you could use a btree and just take it. Full disclosure this was opus 4.6 that wrote this and it passes all tests and I proof read the code and looked like what I would have done so I decided to ship it. Let me know if you want me to make any changes as the AI could have missed something.

Below is the PR description that claude wrote.

## Summary

Adds a pure Go in-memory backend for kine, registered as the `memory://` scheme. State lives entirely in process memory and is lost on shutdown. Intended for ephemeral and test scenarios where the SQL or NATS layers underneath the existing backends are unwanted overhead.

A side effect of building this driver was extracting the lease/TTL machinery out of `pkg/logstructured` into a new `pkg/ttl` package so both backends share one implementation.

## Why this vs. nats memory or sqlite `:memory:`

Both alternatives can keep data in RAM, but they still pay the cost of the layers underneath kine. This driver skips them entirely.

- **vs. nats memory store**: NATS still runs an embedded JetStream server with its own goroutines, KV bucket, JSON encode/decode per entry, and the kine NATS backend layers a btree watcher on top for replay. Even with non-persistent storage you're paying the protocol stack.
- **vs. sqlite `:memory:`**: every operation goes through the SQL parser/planner, the cgo bridge into sqlite3, and kine's `LogStructured` + `SQLLog` path on top of that (poll loop, gap-fill, broadcaster, transaction serialization). `:memory:` only removes the disk; the SQL machinery still runs on every read and write.

This driver is direct in-process state: a slice for the revision log, a btree for key to history, an atomic for the revision counter, a chan-close for watch notification. No SQL, no cgo, no IPC, no encoding.

What that buys you in practice:

- **Faster startup/teardown.** No schema setup, no embedded server, no socket; tests reset by allocating a new `Memory`.
- **No filesystem or port state.** Nothing to clean up between runs, no port allocation conflicts when running tests in parallel.
- **Determinism.** No WAL checkpoints, no async flushes, no compaction lag. Operations land in order and reads see exactly what was written.
- **Smaller surface for test flakes.** Fewer moving parts means fewer things to blame when a test is slow or racy.

Tradeoff: state is process-local and lost on exit. That's the point. This is for tests, ephemeral clusters, and dev loops, not production.

## How the storage is laid out

The driver keeps two structures in sync, and they answer different questions:

- A flat **revision log** (`[]*entry`), one slot per write, ordered by monotonic `currentRevision` (`atomic.Int64`). Source of truth for "what happened, in order." `Watch` replays
it; historical reads at a revision walk it. Each `entry` carries a `prev *entry` pointer to its predecessor for the same key, so previous-KV lookups during watch fanout are O(1)
without re-indexing.
- A **btree** keyed by string (`tidwall/btree.Map[string, []*entry]`) where each value is the full history of that key. Answers "what's at key X" and "what's under prefix P, in
sorted order."

A btree (not a hash map) because the apiserver leans on **ordered range queries**: `List` over a prefix has to return keys in sorted order, and pagination uses a start key to resume mid-range. A hash map gives O(1) point lookup but no order; sorting on every list would be O(n log n) per call. A btree gives both: O(log n) point lookup *and* an ordered iterator you can seek into.

The library is `tidwall/btree`, an in-memory copy-on-write B-tree. Each node holds many keys, which keeps the tree shallow and cache-friendly, comparable to what databases use on disk but tuned for RAM. For this driver:

- **Point lookup** (`Get`, `latest`, `atRevision`): `keys.Get(key)` returns the history slice; we take the last entry, or scan the slice for the largest revision <= a target.
- **Prefix / range scan** (`List`, `Count`): `keys.Iter()` plus `iter.Seek(startKey)` jumps to the first key >= start, then `iter.Next()` walks in sorted order, stopping at the
prefix boundary. Cost is O(log n) to seek plus O(k) for k matching keys, with no scan of unrelated keys.
- **Append on write** (`Create`/`Update`/`Delete`): look up the existing history, append a new `entry`, `keys.Set(key, history)` writes it back, O(log n).

Each value being a *slice of entries* is what makes historical reads cheap: `atRevision(key, rev)` walks just that key's slice (typically short) to find the latest entry at or before `rev`, never touching unrelated keys or the global log.

`Compact` advances `compactRevision`, drops compacted entries from each key's history slice, and trims the flat log; the log stays dense, so the index of the first entry strictly after a revision is just `rev - compactRevision` (clamped). Reads below the boundary return `ErrCompacted`.

## Shared TTL package

`pkg/logstructured/logstructured.go` had an inline lease-expiration implementation: a delaying workqueue seeded from a list, then watching for new lease events and deleting each key when its TTL elapses. The memory driver needs the same behavior, and the only contract is the `List` / `Watch` / `Delete` subset of `server.Backend`, so this PR pulls that machinery into `pkg/ttl` and has both backends call `go ttl.Run(ctx, backend)` from `Start`.

Behavior is unchanged for `LogStructured`; the function signatures and trace logging match what was there before.

## Startup state

`Start` seeds the same two entries the SQL and NATS backends produce on startup so that the apiserver and the rest of kine see a consistent contract:

- `compact_rev_key`: the apiserver-visible compaction marker, normally written by `SQLLog.compactStart`.
- `/registry/health`: the liveness key, normally written by `LogStructured.Start`.

Without these, the apiserver rejects the first read with `illegal resource version from storage: 0`, and starting revision lands at 2, which is what the existing `watcher_test.go` patch in `scripts/test-helpers` already expects (`NewTooLargeResourceVersionError(uint64(102), 2, 0)`).

## What's in it

- **`pkg/drivers/memory/memory.go`**: the driver. Implements `server.Backend` directly (it does not go through `LogStructured`).
  - Append-only revision log with key-history btree as described above; `currentRevision` is an `atomic.Int64`.
  - `Get` / `List` / `Count` / `Watch` / `Compact` / `Create` / `Update` / `Delete` matching the contract the apiserver expects, including historical reads at a revision and
`ErrCompacted` / `ErrFutureRev` semantics.
  - `Watch` uses a chan-close broadcast notified on every append; watchers replay the log from their last-seen revision and filter by prefix. `entry.prev` provides O(1) previous-KV
  access for events.
  - Lease/TTL handled by the shared `pkg/ttl` package.
- **`pkg/drivers/memory/memory_test.go`**: package-level unit tests covering create/get/update/delete, list/range/prefix iteration, historical reads, watches (initial backlog, new events, prev-KV), compaction, and TTL expiration end-to-end.
- **`pkg/ttl/ttl.go`**: new shared package. `Run(ctx, backend)` consumes leased keys and deletes each one when its TTL elapses. Pagination loop terminates correctly whether or not the backend honors `limit` (the in-memory backend ignores it).
- **`pkg/logstructured/logstructured.go`**: deletes the inline TTL implementation and calls `ttl.Run` instead. No behavior change for SQL backends.
- **`pkg/endpoint/init.go`**: registers the driver via blank import so `memory://` resolves at runtime.
- **`pkg/testserver/test_server.go`**: recognizes the `memory` scheme in `setDatabasePath` (no DSN rewrite needed; pass through unchanged) so the upstream etcd3 apiserver test
harness can target it.
- **`scripts/test-run-memory`**: runner that invokes the upstream etcd3 storage tests against `memory://` and then provisions a k3s cluster pointed at a kine container using the in-memory backend, mirroring the existing `test-run-sqlite` / `test-run-nats` runners.

## Test plan

- [x] `go test ./pkg/drivers/memory/...`
- [x] `go test ./pkg/logstructured/...` (TTL extraction is non-breaking)
- [x] `source ./scripts/test-helpers && build-apiserver-tests`
- [x] `KINE_ENDPOINT=memory:// ./bin/etcd3.test -test.failfast -test.count 1 -test.parallel 1 -test.timeout 15m`, all 78 tests pass
- [ ] `scripts/test-run-memory` end-to-end (kine container + k3s cluster)

### DIFF
--- a/pkg/drivers/memory/memory.go
+++ b/pkg/drivers/memory/memory.go
@@ -1,0 +1,535 @@
+package memory
+
+import (
+	"context"
+	"strings"
+	"sync"
+	"sync/atomic"
+
+	"github.com/k3s-io/kine/pkg/drivers"
+	"github.com/k3s-io/kine/pkg/server"
+	"github.com/k3s-io/kine/pkg/ttl"
+	"github.com/sirupsen/logrus"
+	"github.com/tidwall/btree"
+)
+
+func init() {
+	drivers.Register("memory", New)
+}
+
+func New(ctx context.Context, wg *sync.WaitGroup, cfg *drivers.Config) (bool, server.Backend, error) {
+	logrus.Info("using in-memory backend")
+	return false, &Memory{
+		keys:     btree.NewMap[string, []*entry](0),
+		notifyCh: make(chan struct{}),
+	}, nil
+}
+
+type entry struct {
+	revision       int64
+	key            string
+	value          []byte
+	createRevision int64
+	version        int64
+	prevRevision   int64
+	lease          int64
+	created        bool
+	deleted        bool
+	prev           *entry
+}
+
+func (e *entry) toKeyValue() *server.KeyValue {
+	return &server.KeyValue{
+		Key:            e.key,
+		Value:          e.value,
+		CreateRevision: e.createRevision,
+		ModRevision:    e.revision,
+		Version:        e.version,
+		Lease:          e.lease,
+	}
+}
+
+type Memory struct {
+	mu              sync.RWMutex
+	currentRevision atomic.Int64
+	compactRevision int64
+
+	log  []*entry
+	keys *btree.Map[string, []*entry]
+
+	notifyMu sync.Mutex
+	notifyCh chan struct{}
+}
+
+// explicit interface check
+var _ server.Backend = &Memory{}
+
+func (m *Memory) Start(ctx context.Context) error {
+	// Seed the same startup entries that SQL-backed and NATS backends do:
+	//   1. compact_rev_key — written by SQLLog.compactStart; gives the backend
+	//      a non-zero starting revision (apiserver rejects rev=0).
+	//   2. /registry/health — written by LogStructured.Start; the apiserver
+	//      uses it as a liveness probe.
+	m.mu.Lock()
+	rev := m.nextRevision()
+	m.appendEntry(&entry{
+		revision:       rev,
+		key:            "compact_rev_key",
+		createRevision: rev,
+		version:        1,
+		created:        true,
+	})
+	rev = m.nextRevision()
+	m.appendEntry(&entry{
+		revision:       rev,
+		key:            "/registry/health",
+		value:          []byte(`{"health":"true"}`),
+		createRevision: rev,
+		version:        1,
+		created:        true,
+	})
+	m.mu.Unlock()
+
+	go ttl.Run(ctx, m)
+	return nil
+}
+
+func (m *Memory) CurrentRevision(ctx context.Context) (int64, error) {
+	return m.currentRevision.Load(), nil
+}
+
+func (m *Memory) DbSize(ctx context.Context) (int64, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+	var size int64
+	for _, e := range m.log {
+		size += int64(len(e.key) + len(e.value) + 64)
+	}
+	return size, nil
+}
+
+func (m *Memory) WaitForSyncTo(revision int64) {}
+
+// nextRevision allocates and returns the next revision. Increment is
+// atomic; callers still need m.mu (write) when pairing the new revision
+// with an appendEntry, so the log stays in revision order.
+func (m *Memory) nextRevision() int64 {
+	return m.currentRevision.Add(1)
+}
+
+// appendEntry adds e to the log and to its key's history, wiring up
+// e.prev to point at the previous entry for the same key (if any).
+// Caller must hold m.mu (write).
+func (m *Memory) appendEntry(e *entry) {
+	m.log = append(m.log, e)
+	hist, _ := m.keys.Get(e.key)
+	if len(hist) > 0 {
+		e.prev = hist[len(hist)-1]
+	}
+	m.keys.Set(e.key, append(hist, e))
+}
+
+// broadcast wakes every goroutine currently blocked on getNotifyCh.
+// Implemented as close-and-replace rather than sync.Cond.Broadcast so that
+// waiters can select on ctx.Done() alongside the wake — sync.Cond.Wait()
+// can't be combined with context cancellation without a helper goroutine.
+func (m *Memory) broadcast() {
+	m.notifyMu.Lock()
+	close(m.notifyCh)
+	m.notifyCh = make(chan struct{})
+	m.notifyMu.Unlock()
+}
+
+// getNotifyCh returns a channel that will be closed on the next broadcast.
+// Watchers should grab it BEFORE scanning the log so they don't miss a
+// broadcast that fires while they're iterating.
+func (m *Memory) getNotifyCh() <-chan struct{} {
+	m.notifyMu.Lock()
+	ch := m.notifyCh
+	m.notifyMu.Unlock()
+	return ch
+}
+
+// latest returns the most recent entry for key, or nil if none exists.
+// The returned entry may be a tombstone (deleted == true).
+// Caller must hold m.mu (read or write).
+func (m *Memory) latest(key string) *entry {
+	hist, ok := m.keys.Get(key)
+	if !ok || len(hist) == 0 {
+		return nil
+	}
+	return hist[len(hist)-1]
+}
+
+// atRevision returns the entry for key with the highest revision <= revision,
+// or nil if no such entry exists.
+// Caller must hold m.mu (read or write).
+func (m *Memory) atRevision(key string, revision int64) *entry {
+	hist, ok := m.keys.Get(key)
+	if !ok {
+		return nil
+	}
+	var result *entry
+	for _, e := range hist {
+		if e.revision <= revision {
+			result = e
+		} else {
+			break
+		}
+	}
+	return result
+}
+
+// logIndexAfter returns the index of the first log entry with revision > rev.
+// The log is dense and ordered, with m.log[0].revision == m.compactRevision+1,
+// so the index is just rev - m.compactRevision (clamped). For rev values at or
+// below m.compactRevision the result is 0; for rev at or above the highest
+// stored revision the result is len(m.log).
+// Caller must hold m.mu (read or write).
+func (m *Memory) logIndexAfter(rev int64) int {
+	i := rev - m.compactRevision
+	if i < 0 {
+		i = 0
+	}
+	if i > int64(len(m.log)) {
+		i = int64(len(m.log))
+	}
+	return int(i)
+}
+
+// Get returns the current revision and the KeyValue for the given key.
+func (m *Memory) Get(ctx context.Context, key, rangeEnd string, limit, revision int64, keysOnly bool) (int64, *server.KeyValue, error) {
+	if strings.HasSuffix(key, "/") && rangeEnd == "" {
+		key = key[:len(key)-1]
+	}
+	rev, kvs, err := m.List(ctx, key, rangeEnd, limit, revision, keysOnly)
+	if err != nil {
+		return rev, nil, err
+	}
+	if len(kvs) == 0 {
+		return rev, nil, nil
+	}
+	return rev, kvs[0], nil
+}
+
+func (m *Memory) Create(ctx context.Context, key string, value []byte, lease int64) (int64, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	latest := m.latest(key)
+	if latest != nil && !latest.deleted {
+		return m.currentRevision.Load(), server.ErrKeyExists
+	}
+
+	rev := m.nextRevision()
+	var prevRev int64
+	if latest != nil {
+		prevRev = latest.revision
+	}
+
+	m.appendEntry(&entry{
+		revision:       rev,
+		key:            key,
+		value:          value,
+		createRevision: rev,
+		version:        1,
+		prevRevision:   prevRev,
+		lease:          lease,
+		created:        true,
+	})
+	m.broadcast()
+	return rev, nil
+}
+
+func (m *Memory) Update(ctx context.Context, key string, value []byte, revision, lease int64) (int64, *server.KeyValue, bool, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	latest := m.latest(key)
+	if latest == nil || latest.deleted {
+		return m.currentRevision.Load(), nil, false, nil
+	}
+	if latest.revision != revision {
+		return m.currentRevision.Load(), latest.toKeyValue(), false, nil
+	}
+
+	rev := m.nextRevision()
+	e := &entry{
+		revision:       rev,
+		key:            key,
+		value:          value,
+		createRevision: latest.createRevision,
+		version:        latest.version + 1,
+		prevRevision:   latest.revision,
+		lease:          lease,
+	}
+	m.appendEntry(e)
+	m.broadcast()
+	return rev, e.toKeyValue(), true, nil
+}
+
+func (m *Memory) Delete(ctx context.Context, key string, revision int64) (int64, *server.KeyValue, bool, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	latest := m.latest(key)
+	if latest == nil || latest.deleted {
+		return m.currentRevision.Load(), nil, false, nil
+	}
+	if revision != 0 && latest.revision != revision {
+		return m.currentRevision.Load(), latest.toKeyValue(), false, nil
+	}
+
+	rev := m.nextRevision()
+	m.appendEntry(&entry{
+		revision:       rev,
+		key:            key,
+		value:          latest.value,
+		createRevision: latest.createRevision,
+		version:        latest.version,
+		prevRevision:   latest.revision,
+		lease:          latest.lease,
+		deleted:        true,
+	})
+	m.broadcast()
+	return rev, latest.toKeyValue(), true, nil
+}
+
+func (m *Memory) List(ctx context.Context, prefix, startKey string, limit, revision int64, keysOnly bool) (int64, []*server.KeyValue, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	rev := m.currentRevision.Load()
+	if revision > 0 {
+		if revision > rev {
+			return rev, nil, server.ErrFutureRev
+		}
+		if revision < m.compactRevision {
+			return rev, nil, server.ErrCompacted
+		}
+		rev = revision
+	}
+
+	seek, exact := seekKey(prefix, startKey)
+	iter := m.keys.Iter()
+	if !iter.Seek(seek) {
+		return rev, nil, nil
+	}
+
+	var kvs []*server.KeyValue
+	for {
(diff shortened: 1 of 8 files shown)


brandond:
👋🏻 @luthermonson - let me take a shot at reviewing this.

pkg/drivers/memory/memory.go
```diff
@@ -0,0 +1,543 @@
...
+	b.expireMu.Unlock()
+	select {
+	case b.expireWake <- struct{}{}:
+	default:
+	}
+}
+
+func (b *Backend) expireLoop(ctx context.Context) {
+	var timer *time.Timer
+	var timerC <-chan time.Time
```
brandond: This looks like a mashup of code reused from other places:
* https://github.com/k3s-io/kine/blob/648a2daa906f4c25efd83e5dd20af0e69888fae8/pkg/util/wait.go#L13
* https://github.com/k3s-io/kine/blob/648a2daa906f4c25efd83e5dd20af0e69888fae8/pkg/util/wait.go#L59
* https://github.com/k3s-io/kine/blob/648a2daa906f4c25efd83e5dd20af0e69888fae8/pkg/logstructured/logstructured.go#L288

Can you refactor this to use existing utility code and function more like the logstructured code, which uses a delaying queue? The code here directly invokes `scheduleExpire` from Create/Update/Delete; I'd prefer to see the TTL workqueue just feed from a watch same as the other implementations.

pkg/drivers/memory/memory.go
```diff
@@ -0,0 +1,543 @@
...
+	return rev, nil
+}
+
+func (b *Backend) Update(ctx context.Context, key string, value []byte, revision, lease int64) (int64, *server.KeyValue, bool, error) {
+	b.mu.Lock()
+
+	latest := b.latest(key)
+	if latest == nil || latest.deleted {
+		rev := b.currentRevision
+		b.mu.Unlock()
```
brandond: can we just `defer b.mu.Unlock()` after locking, instead of explicitly unlocking at every individual return site? I don't see any paths where we would NOT want to unlock when returning.

pkg/drivers/memory/memory.go
```diff
@@ -0,0 +1,543 @@
...
+		watchRev = rev
+	}
+
+	return server.WatchResult{
+		CurrentRevision: watchRev,
+		Events:          events,
+	}
+}
+
+func (b *Backend) Compact(ctx context.Context, revision int64) (int64, error) {
```
brandond: Am I correct in understanding that this does not actually support compacting, and will grow without bound? Is that a problem? I suspect it might reduce the usability of this backend for longer-running tests, or tests on nodes with less resources - which is exactly where this would be most useful.

Have you explored how much trouble it might be to actually implement this? I bet it wouldn't be too hard, if you replaced direct access to `b.log` with some helpers that also handled min/max rev and locking.

pkg/drivers/memory/memory.go
```diff
@@ -0,0 +1,543 @@
...
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if revision > b.currentRevision {
+		return b.currentRevision, nil
+	}
+	b.compactRevision = revision
+	return b.currentRevision, nil
+}
+
+func seekKey(prefix, startKey string) (string, bool) {
```
brandond, line 523: Can you provide a doc comment on what is going on here, and what's returned?

pkg/drivers/memory/memory.go
```diff
@@ -0,0 +1,543 @@
...
+func (h expireHeap) Swap(i, j int)        { h[i], h[j] = h[j], h[i] }
+func (h *expireHeap) Push(x any)          { *h = append(*h, x.(*expireEntry)) }
+func (h *expireHeap) Pop() any {
+	old := *h
+	n := len(old)
+	e := old[n-1]
+	old[n-1] = nil
+	*h = old[:n-1]
+	return e
+}
```
brandond: this appears to be all related to the TTL work queue, and can be removed if we just lean on `workqueue.NewTypedDelayingQueue` as discussed above.

pkg/drivers/memory/memory.go
luthermonson: great feedback, ill get this fixed

pkg/drivers/memory/memory.go
luthermonson: ya that would look cleaner, ill get this in too

pkg/drivers/memory/memory.go
luthermonson: i havent but ill give it some time, was trying to avoid anything that looked like you were gonna keep using this long term but i dont mind trying to get it working

pkg/drivers/memory/memory.go
luthermonson: the rewrite ended up looking a lot better, claude even fixed the returns so the func is overall more readible

pkg/drivers/memory/memory.go
luthermonson: the work queue rewrite looked pretty successful, resolving since this is ripped out

pkg/drivers/memory/memory.go
luthermonson: rewrite on this looks functional, and claude added 4 unit tests to validate the functionality. pretty sure we're good here now.

pkg/drivers/memory/memory.go
luthermonson: added a unit test for TTL validation as well, this rewrite came out pretty good. thanks for the suggestion.

luthermonson:
i think everything was addressed, separate commits for now for the new changes but if you want it all squashed into one before you merge let me know

pkg/drivers/memory/memory.go
```diff
@@ -0,0 +1,648 @@
...
+
+func (b *Backend) appendEntry(e *entry) {
+	b.log = append(b.log, e)
+	hist, _ := b.keys.Get(e.key)
+	if len(hist) > 0 {
+		e.prev = hist[len(hist)-1]
+	}
+	b.keys.Set(e.key, append(hist, e))
+}
```
brandond: maybe note that these should only be called while already holding a write lock?

pkg/drivers/memory/memory.go
```diff
@@ -0,0 +1,648 @@
...
+
+func (b *Backend) broadcast() {
+	b.notifyMu.Lock()
+	close(b.notifyCh)
+	b.notifyCh = make(chan struct{})
+	b.notifyMu.Unlock()
+}
```
brandond: Is this a hand-rolled version of [Broadcast](https://pkg.go.dev/sync#Cond.Broadcast) from sync.Cond? Why?

pkg/drivers/memory/memory.go
```diff
@@ -0,0 +1,648 @@
...
+		size += int64(len(e.key) + len(e.value) + 64)
+	}
+	return size, nil
+}
+
+func (b *Backend) WaitForSyncTo(revision int64) {}
+
+func (b *Backend) nextRevision() int64 {
+	b.currentRevision++
+	return b.currentRevision
```
brandond: I feel like if you made this an `atomic.Int64` you could just use `Add()` to increment and `Load()` to read without having to manage locking it by hand...

pkg/drivers/memory/memory.go
```diff
@@ -0,0 +1,648 @@
...
+	return result
+}
+
+// logIndexAfter returns the index of the first log entry with revision > rev.
+// Caller must hold at least a read lock on b.mu.
+func (b *Backend) logIndexAfter(rev int64) int {
+	return sort.Search(len(b.log), func(i int) bool {
+		return b.log[i].revision > rev
+	})
+}
```
brandond: Shouldn't compaction always trim all log entries up to compactRev? If this is true then you should be able to perform this operation with basic math - ie `i = rev - compactRev`, since the log is by definition ordered by revision. If `rev < compactRev` then the event has been compacted and no longer exists.

pkg/drivers/memory/memory.go
```diff
@@ -0,0 +1,648 @@
...
+	filtered := b.log[:0]
+	for _, e := range b.log {
+		if _, ok := keep[e]; ok {
+			filtered = append(filtered, e)
+		}
+	}
+	for i := len(filtered); i < len(b.log); i++ {
+		b.log[i] = nil
+	}
+	b.log = filtered
```
brandond: I am confused as to why you need to filter it like this; by definition all revisions older than compactRev should be inaccessible. You can still get the key from the KV store but the revision event itself is gone from the log. Why can't you just do `b.log = b.log[revision:]` ?

brandond (commented):
couple more nits/questions

pkg/drivers/memory/memory.go
```diff
@@ -0,0 +1,648 @@
...
+	})
+}
+
+// ttl runs a long-lived goroutine that watches the backend for entries with
+// a non-zero Lease and schedules them for deletion once their TTL expires.
+// It mirrors logstructured.LogStructured.ttl: an initial list seeds the
+// delaying workqueue with any pre-existing leased keys, then a watch picks
+// up new ones. A handler goroutine consumes the queue and calls Delete when
+// each entry reaches its expiration time.
+func (b *Backend) ttl(ctx context.Context) {
```
brandond: I feel like maybe this/handleTTLEvent/loadTTLEventKV/storeTTLEventKV (from both here and logstructured) could all go into shared ttl package now? They just need a handle to the backend to call List/Watch/Delete which are not backend-specific.

pkg/drivers/memory/memory.go
```diff
@@ -0,0 +1,648 @@
...
+var _ server.Backend = (*Backend)(nil)
+
+func NewBackend() *Backend {
+	return &Backend{
+		keys:     btree.NewMap[string, []*entry](0),
+		notifyCh: make(chan struct{}),
+	}
+}
+
+func (b *Backend) Start(ctx context.Context) error {
```
brandond: nit: Backend is the interface that this satisfies; name the struct after what it is, not the interface it implements. I realize we let the NATS backend get away with calling their struct `Backend` too but that should probably be fixed.

```suggestion
type Memory struct {
	mu              sync.RWMutex
	currentRevision int64
	compactRevision int64

	log  []*entry
	keys *btree.Map[string, []*entry]

	notifyMu sync.Mutex
	notifyCh chan struct{}
}

// explicit interface check
var _ server.Backend = &Memory{}

func (m *Memory) Start(ctx context.Context) error {
```

pkg/drivers/memory/memory.go
```diff
@@ -0,0 +1,648 @@
...
+const ttlRetryInterval = 250 * time.Millisecond
+
+func init() {
+	drivers.Register("memory", New)
+}
+
+func New(ctx context.Context, wg *sync.WaitGroup, cfg *drivers.Config) (bool, server.Backend, error) {
+	logrus.Info("using in-memory backend")
+	return false, NewBackend(), nil
+}
```
brandond, line 26: This does nothing but call NewBackend, which isn't used anywhere else and only creates a new instance of the struct. Just move all the logic into one place.

```suggestion
func New(ctx context.Context, wg *sync.WaitGroup, cfg *drivers.Config) (bool, server.Backend, error) {
	logrus.Info("using in-memory backend")
	memory := &Memory{
		keys:     btree.NewMap[string, []*entry](0),
		notifyCh: make(chan struct{}),
	}
	return false, memory , nil
}
```

brandond:
I also need to fix the golangci-lint config so that it doesn't complain about `wg.Add()`:
* https://github.com/k3s-io/kine/pull/654

pkg/drivers/memory/memory.go
luthermonson: Have you met Claude?

pkg/drivers/memory/memory.go
luthermonson: it would be on all appendEntry, nextRevision, latest, atRevision, and broadcast? no problem to add comments

pkg/drivers/memory/memory.go
luthermonson: not against, but feels like a separate PR. you sure you want to scope creep this?

pkg/drivers/memory/memory.go
luthermonson: ya this is a worthwhile change, ill get this in

pkg/drivers/memory/memory.go
luthermonson: this did real work in the first implementation of Compact but your concerns will be resolved with the rewrite of Compact. I'm going to walk the changes in and that will be the last one. I'm just resolving this comment now as it wont be valid after new code comes in.

pkg/drivers/memory/memory.go
luthermonson: a little cleaner with same outcomes. i went through the lock and increment logic and it is all correct but atomic.Int64 makes it more idiomatic go. ill get this changed.

pkg/drivers/memory/memory.go
luthermonson: claude admitted to me that it only kept the floors idea because there was an existing test that didnt care about compaction before it was added and it was trying to fulfill the contract of the test rather than change the test. i will get this switched out since it makes more sense.

pkg/drivers/memory/memory.go
luthermonson: I went through both options and I think we should stick with the channel pattern as it gives you more control over things you would need to implement on the side and add complexity for instance if the context is cancelled...

1. Spawn a helper goroutine that calls Broadcast when ctx is canceled. The waiter wakes, checks ctx.Err(), exits.
2. Poll — wake periodically from Wait() (e.g. with time.AfterFunc poking the cond) and check ctx.Err().

All seems unnecessary when we have a channel implementation that solves everything we need without the sync.Cond being introduced.

pkg/drivers/memory/memory.go
brandond: Yeah, if it is reusable by multiple backends, I'm ok moving it to a new package here, rather than temporarily duplicating it.

luthermonson: as you wish

luthermonson: commit up, would you also like to migrate NATS to the ttl package? it's doable but needs a bit more work which is not really out of scope for this PR. willing to do it though if you want it now

luthermonson: here's a run at migrating NATs if you want me to cherry pick this into this PR i can https://github.com/luthermonson/kine/compare/memory-driver...luthermonson:kine:nats-ttl-migration?expand=1

brandond:
rebased and addressed revive check so that we can see how the rest of the tests look

pkg/drivers/memory/memory.go
brandond: We can leave the nats stuff out of scope for now, I tend to let the synadia folks handle it.

pkg/drivers/memory/memory.go
brandond: idk it still feels a little weird, I guess we can revisit it later if I think of a tidier way to handle this.

brandond (approved):
LGTM. I think I'll probably squash when merging.

luthermonson:
squashed with one nice big commit message that is current with the changes we made.

PR #660, opened 2026-05-04
Bump to Kubernetes 1.35

brandond:
No API changes, so this will not break compat with older minors of K3s.

Not bumping to 1.36 yet as that requires go1.26


brandond:
hmm, looks like nats needs more work :(

brandond:
Appears nats does not properly handle keys with trailing slashes... the delete transaction here fails on nats but works on other backends:

```bash
etcdctl put /pods/ foo
etcdctl get /pods/
echo -e '\nget /pods/\ndel /pods/\n\n' | etcdctl txn
etcdctl get /pods/
```

Store/list/watch/delete of keys with trailing slashes is now explicitly tested for:
https://github.com/kubernetes/kubernetes/blob/v1.35.4/staging/src/k8s.io/apiserver/pkg/storage/testing/store_tests.go#L4505-L4507

The same test fails earlier for memory - it is not possible to get a key with a trailing slash as tested here: 
https://github.com/kubernetes/kubernetes/blob/v1.35.4/staging/src/k8s.io/apiserver/pkg/storage/testing/store_tests.go#L4443

vitorsavian (approved):

mgfritch (approved):

PR #663, opened 2026-05-08
Update to Kubernetes 1.36 / golang 1.26

brandond:


dereknola:
This still relevant @brandond ?

brandond:
I was going to leave it open until we are no longer using go1.25 in any of our k3s release branches, at which point we can merge it. The alternative would be branching kine for older minors and that seems like too much work.

brandond:
Closing in favor of https://github.com/k3s-io/kine/pull/739

Issue #651, opened 2026-04-24
listSQL causes Using temporary; Using filesort on MySQL/Aurora leading to slow queries under concurrent load


rajeshkio:
<p><strong>Description</strong></p>
<p>When using KINE with a MySQL or Aurora MySQL backend, the <code>listSQL</code> query in <code>pkg/drivers/generic/generic.go</code> produces a query plan that includes <code>Using temporary; Using filesort</code> on every execution. In isolation this is acceptable, but under concurrent load — which is normal during Kubernetes watch/list operations — this causes severe InnoDB contention that degrades query performance by 125-350x, consistently exceeding KINE's 1-second slow query threshold and triggering cascading watch timeouts.</p>
<hr>
<p><strong>Environment</strong></p>
<ul>
<li>KINE version: current master</li>
<li>Database: Aurora Serverless v2 MySQL 8.0 (reproduced on <code>stg-rajesh-tenant-cluster</code>, <code>us-west-2</code>)</li>
<li>k3s version: k3s v1.x (journal logs from node <code>ip-10-10-100-155</code>)</li>
<li>Table size: ~5,775 rows (small — this is not a table size problem)</li>
<li>Aurora MinCapacity: 1.5 ACU</li>
</ul>
<hr>
<p><strong>The Problem</strong></p>
<p>The current <code>listSQL</code> in <code>generic.go</code>:</p>
<pre><code class="language-go">listSQL = fmt.Sprintf(`
    SELECT (%s), (%s), %s
    FROM kine AS kv
    JOIN (
        SELECT MAX(mkv.id) AS id
        FROM kine AS mkv
        WHERE
            mkv.name LIKE ?
            %%s
        GROUP BY mkv.name) maxkv
    ON maxkv.id = kv.id
    WHERE
        (kv.deleted = 0 OR ?)
    ORDER BY kv.id ASC
    `, revSQL, compactRevSQL, columns)
</code></pre>
<p>This generates a query that MySQL cannot optimize because the <code>GROUP BY</code> subquery forces materialization of a derived table. The outer <code>ORDER BY</code> then has to sort this derived table result with no index available.</p>
<p><strong>EXPLAIN output (current query):</strong></p>
<pre><code>+----+-------------+------------+--------+------+----------+----------------------------------------------+
| id | select_type | table      | type   | rows | filtered | Extra                                        |
+----+-------------+------------+--------+------+----------+----------------------------------------------+
[...]

PR #668, opened 2026-05-14
Improve SQL query performance and metrics

brandond:
### Proposed Changes

* Reorders filter expressions and adds optimized direct key query to reduce use of subqueries and temp tables.
* Uses a prepared statement for watch polling, instead of re-evaluating it again every time.
* Uniquely names each query, and uses the query name as a label in latency metrics and log messages
* Reworks postgres queries to better align with other engines.
* Increases default max-idle-conns to 20 to avoid constant connection thrashing under even moderate load
* Adds a SQL server hostname cache to reduce delays caused by excessive DNS lookups when operating with an undersized connection pool.
* Fixes several race conditions where the reported current revision could temporarily go backwards under heavy load. No data was lost, but clients could see incorrect `required revision is a future revision` errors.

Unfortunately is is not possible to eliminate the filesort on name, although the query cost is still reduced by half due to elimination of the inner join.

### Linked Issues:
* https://github.com/k3s-io/kine/issues/651

### Additional Notes

* Moving the `ORDER BY name` clause into the subquery (to eliminate the filesort) works some of the time, but the order of rows across an inner join is unspecified so it cannot be relied upon. In practice it does seem to work most of the time, but not when a `LIMIT` is in effect as the engine will limit the number of rows produced by the subquery so that later revisions are not seen, producing incorrect results.
* `SELECT ... FROM kine WHERE id IN (SELECT MAX(id) FROM kine WHERE name LIKE ? ESCAPE '^' AND name >= ? GROUP BY name)` is less expensive when the list of IDs is large, but more expensive when the list of IDs is small - and TERRIBLE when only a single ID is matched.
* `SELECT ... FROM kine INNER JOIN (SELECT MAX(id) AS id FROM kine WHERE name LIKE ? ESCAPE '^' AND name >= ? GROUP BY name) AS mkv USING (id)` is slightly more expensive when the list of IDs is large, but less expensive when the list is short.
* Queries that do not need any matching behavior - just the latest ID for a specific name - can be optimized since we know that there will be only a single ID.

This PR also enhances slow SQL warnings to show a redacted version of the query with args filled:
> `INFO[2026-05-20T19:31:56.421415989Z] Slow SQL: INSERT INTO kine(name, created, deleted, create_revision, prev_revision, lease, value, old_value) values('/registry/health', 1, 0, 0, 0, 0, [17]byte(...), [0]byte(...))  duration=5.868734ms name=InsertLastInsertID started="2026-05-20T19:31:56.415526682Z"`
> `INFO[2026-05-20T19:31:58.424805073Z] Slow SQL: SELECT current_rev, compact_rev, id, name, created, deleted, create_revision, prev_revision, lease, value, old_value FROM (SELECT MAX(id) AS current_rev FROM kine) AS current, (SELECT MAX(prev_revision) AS compact_rev FROM kine WHERE name = 'compact_rev_key') AS compact, kine WHERE name LIKE '%' ESCAPE '^' AND id > 2 ORDER BY id ASC LIMIT 500  duration="644.935µs" name=AfterOldVal started="2026-05-20T19:31:58.424126367Z"`

### Performance Comparison

CI runs a very lightweight load test that just churns configmaps; this shows a roughly 20% improvement in average request latencies across all operation types. CPU and memory utilization are not measured.

```
BEFORE
[mysql-8.4] [PERF]	0.004  average mysql-8.4 request duration (seconds): create configmaps
[mysql-8.4] [PERF]	0.005  [899]  ██████████████████████████████████████████████████
[mysql-8.4] [PERF]	0.025  [239]  █████████████▎
[mysql-8.4] [PERF]	0.05   [  0]
--
[mysql-8.4] [PERF]	0.004  average mysql-8.4 request duration (seconds): delete configmaps
[mysql-8.4] [PERF]	0.005  [310]  ██████████████████████████████████████████████████
[mysql-8.4] [PERF]	0.025  [212]  ██████████████████████████████████▎
[mysql-8.4] [PERF]	0.05   [  3]  ▌
--
[mysql-8.4] [PERF]	0.002  average mysql-8.4 request duration (seconds): get configmaps
[mysql-8.4] [PERF]	0.005  [3337]  ██████████████████████████████████████████████████
[mysql-8.4] [PERF]	0.025  [ 140]  ██▏
[mysql-8.4] [PERF]	0.05   [   0]
--
[mysql-8.4] [PERF]	0.000  average mysql-8.4 request duration (seconds): list configmaps
[mysql-8.4] [PERF]	0.005  [1]  ██████████████████████████████████████████████████
[mysql-8.4] [PERF]	0.025  [0]
[mysql-8.4] [PERF]	0.05   [0]
--
[mysql-8.4] [PERF]	0.005  average mysql-8.4 request duration (seconds): update configmaps
[mysql-8.4] [PERF]	0.005  [873]  ██████████████████████████████████████████████████
[mysql-8.4] [PERF]	0.025  [780]  ████████████████████████████████████████████▋
[mysql-8.4] [PERF]	0.05   [  1]

AFTER
[mysql-8.4] [PERF]	0.003  average mysql-8.4 request duration (seconds): create configmaps
[mysql-8.4] [PERF]	0.005  [1009]  ██████████████████████████████████████████████████
[mysql-8.4] [PERF]	0.025  [ 101]  █████
[mysql-8.4] [PERF]	0.05   [   1]
--
[mysql-8.4] [PERF]	0.004  average mysql-8.4 request duration (seconds): delete configmaps
[mysql-8.4] [PERF]	0.005  [457]  ██████████████████████████████████████████████████
[mysql-8.4] [PERF]	0.025  [ 91]  ██████████
[mysql-8.4] [PERF]	0.05   [  1]  ▏
--
[mysql-8.4] [PERF]	0.001  average mysql-8.4 request duration (seconds): get configmaps
[mysql-8.4] [PERF]	0.005  [3462]  ██████████████████████████████████████████████████
[mysql-8.4] [PERF]	0.025  [  79]  █▏
[mysql-8.4] [PERF]	0.05   [   2]
--
[mysql-8.4] [PERF]	0.000  average mysql-8.4 request duration (seconds): list configmaps
[mysql-8.4] [PERF]	0.005  [0]
[mysql-8.4] [PERF]	0.025  [1]  ██████████████████████████████████████████████████
[mysql-8.4] [PERF]	0.05   [0]
--
[mysql-8.4] [PERF]	0.004  average mysql-8.4 request duration (seconds): update configmaps
[mysql-8.4] [PERF]	0.005  [1189]  ██████████████████████████████████████████████████
[mysql-8.4] [PERF]	0.025  [ 514]  █████████████████████▋
[mysql-8.4] [PERF]	0.05   [   3]  ▏


BEFORE
[postgres-17] [PERF]	0.004  average postgres-17 request duration (seconds): create configmaps
[postgres-17] [PERF]	0.005  [842]  ██████████████████████████████████████████████████
[postgres-17] [PERF]	0.025  [288]  █████████████████▏
[postgres-17] [PERF]	0.05   [  7]  ▍
--
[postgres-17] [PERF]	0.004  average postgres-17 request duration (seconds): delete configmaps
[postgres-17] [PERF]	0.005  [324]  ██████████████████████████████████████████████████
[postgres-17] [PERF]	0.025  [201]  ███████████████████████████████
[postgres-17] [PERF]	0.05   [  3]  ▌
--
[postgres-17] [PERF]	0.003  average postgres-17 request duration (seconds): get configmaps
[postgres-17] [PERF]	0.005  [2974]  ██████████████████████████████████████████████████
[postgres-17] [PERF]	0.025  [ 518]  ████████▊
[postgres-17] [PERF]	0.05   [  20]  ▍
--
[postgres-17] [PERF]	0.000  average postgres-17 request duration (seconds): list configmaps
[postgres-17] [PERF]	0.005  [0]
[postgres-17] [PERF]	0.025  [0]
[postgres-17] [PERF]	0.05   [1]  ██████████████████████████████████████████████████
--
[postgres-17] [PERF]	0.006  average postgres-17 request duration (seconds): update configmaps
[postgres-17] [PERF]	0.005  [986]  ██████████████████████████████████████████████████
[postgres-17] [PERF]	0.025  [681]  ██████████████████████████████████▌
[postgres-17] [PERF]	0.05   [  9]  ▌

AFTER
[postgres-17] [PERF]	0.002  average postgres-17 request duration (seconds): create configmaps
[postgres-17] [PERF]	0.005  [1142]  ██████████████████████████████████████████████████
[postgres-17] [PERF]	0.025  [  36]  █▋
[postgres-17] [PERF]	0.05   [   0]
--
[postgres-17] [PERF]	0.002  average postgres-17 request duration (seconds): delete configmaps
[postgres-17] [PERF]	0.005  [500]  ██████████████████████████████████████████████████
[postgres-17] [PERF]	0.025  [ 41]  ████▏
[postgres-17] [PERF]	0.05   [  1]  ▏
--
[postgres-17] [PERF]	0.001  average postgres-17 request duration (seconds): get configmaps
[postgres-17] [PERF]	0.005  [3488]  ██████████████████████████████████████████████████
[postgres-17] [PERF]	0.025  [  96]  █▍
[postgres-17] [PERF]	0.05   [   1]
--
[postgres-17] [PERF]	0.000  average postgres-17 request duration (seconds): list configmaps
[postgres-17] [PERF]	0.005  [0]
[postgres-17] [PERF]	0.025  [1]  ██████████████████████████████████████████████████
[postgres-17] [PERF]	0.05   [0]
--
[postgres-17] [PERF]	0.003  average postgres-17 request duration (seconds): update configmaps
[postgres-17] [PERF]	0.005  [1511]  ██████████████████████████████████████████████████
[postgres-17] [PERF]	0.025  [ 190]  ██████▎
[postgres-17] [PERF]	0.05   [   1]
```

These charts capture two runs - the first segment is Kine v0.15.0, the second is from this PR. Note in particular the drop in connection rate, query latency, and network throughput.
image


pkg/dialer/dialer.go
```diff
@@ -0,0 +1,78 @@
...
+			return nil, err
+		}
+		if len(addrs) == 0 {
+			return nil, &net.AddrError{Err: "LookupHost returned empty address list", Addr: host}
+		}
+		entry = &cacheEntry{
+			addrs:   addrs,
+			expires: time.Now().Add(TTL),
+		}
+		d.cache[host] = entry
```
manuelbuil: isn't there a risk that two concurrent goroutines add an entry at the same time and corrupt the cache?

brandond: Maps are safe for concurrent write, as long as you're not iterating over them at the same time. And it is not possible to "corrupt" a map.

Worst case scenario, multiple goroutines do the lookup at the same time, and all but one's work is wasted.

Also, kine only connects to a single database; other than someone running a custom version of Kine that has somehow been modified to connect to multiple databases the "cache" will only ever contain a single host anyways. Using a map here is probably overkill, but I wanted to keep it generic since we're plugging in at the dialer level.

manuelbuil: I skimmed through an [article](https://medium.com/@diveshkumarchordia/go-maps-and-concurrency-why-you-need-synchronization-and-how-to-get-it-right-6cf5cae2f26b) that explained that maps were not concurrency safe and recommended a new type for that: "sync.Map". Now I see that it does not get corrupted but panics to avoid that corruption if there is a read/write at the same time.

Anyway, I am not super familiar with Kine architecture. So if it is going to be only a single host, there is indeed no concurrency risk

brandond: I just changed it to store a single cache entry with a hostname, instead of using a map.

manuelbuil (approved):
lgtm! (from a non-kine expert)

PR #669, opened 2026-05-17
Add t4 backend driver

makhov:
Adds a new optional backend driver `t4://` that embeds [t4](https://github.com/t4db/t4), an open-source key-value database built on object storage for durable infrastructure state, into kine.

t4 gives kine users a clustered backend that needs no quorum service and no external database — only an S3-compatible bucket (AWS S3, MinIO, Ceph, GCS via S3 API, etc.). Each node keeps a local Pebble DB; the leader uploads checkpoints and streams WAL segments to S3 on a schedule; followers replicate from the leader over gRPC. Leader election is done by writing a lock object to S3, so there is no Raft / no quorum / no split-brain mediation service.

A node that loses its local disk re-bootstraps from the S3 checkpoint + WAL on next start with no operator intervention.

## Why this is useful for kine

- **k3s / k0s/ another lightweight Kubernetes on a handful of nodes.** sqlite is the current "no external deps" option but is single-node; t4 keeps the "no external DB" property while adding HA and offsite durability.
- **Edge / air-gapped-ish deployments with a nearby object store.** Many edge environments already run MinIO or a similar S3 API for image and artifact storage. Pointing kine at the same bucket gives durable HA control-plane state without standing up a separate database tier.
- **Backups are built-in.** Every checkpoint and WAL segment is already in S3 as part of normal operation — no separate backup job, no etcd snapshot cron, no DB dump. Object-store lifecycle rules and versioning give you retention and point-in-time recovery for free.
- **Easy [auto]scaling.** leader election orchestration is done via S3 API, which allows t4 scale as simple as stateless deployment.
- **Disaster recovery is a bucket restore.** Losing all kine nodes and bringing them back from S3 is the documented happy path, not a hand-rolled procedure.
- **Performance.** We in the k0smotron team benchmarked Kubernetes backends ([blog post][bench]) and t4 came out as the highest-throughput kine backend tested — and by a wide margin the most performant embedded one. Throughput and watch behaviour both held up where sqlite and the other embedded options stalled.

[bench]: https://blog.k0sproject.io/posts/k0smotron-big-bang-benchmark/#t4-throughput-and-watch-behavior-at-c1000


brandond:
S3 is an object store served over http. It is not a lock server, it is not a low-latency key-value store. It does not support streaming or watching. While I don't doubt that you could use it as the backend for a generic KV store, to say that I have doubts about its ability to provide the required throughput and latencies required by a production Kubernetes cluster would be putting it mildly.

> followers replicate from the leader over gRPC.

ok, so this requires direct connectivity between cluster members. I am not sure we have another backend with this requirement; generally kine acts as a client only.

Have you actually tried this with a multi-node Kubernetes cluster? If so, against what S3-compliant stores? Is it stable, and how does it scale?

I am concerned that even if we do accept this, we would be taking on maintenance of a backend that sees no actual use due to poor real-world performance. Remember that this projects goal is to provide an etcd-like store for Kubernetes, not for general purpose use.

Kine already supports litestream for replicating to S3, which is I believe a much more appropriate use of an object store. Have you evaluated that?

makhov:
Hi @brandond, thanks for the questions!

Quick context: I'm on the k0s / k0smotron team and I've been running t4 in production as the backend for HA hosted Kubernetes control planes, against AWS S3, Cloudflare R2 and MinIO. As I mentioned in the PR description, we ran a [benchmark](https://blog.k0sproject.io/posts/k0smotron-big-bang-benchmark/#t4-throughput-and-watch-behavior-at-c1000) (including this particular kine driver implementation) and t4 came out as one of the highest-throughput backends tested.

Apiserver reads/writes hit a local Pebble DB, watches are served from that same local DB, and replication between nodes is the leader streaming WAL over gRPC to followers in real time. S3 is used for async checkpoint + WAL archival (background), one-shot leader-election PUT, and cold-start recovery — none of it in any request's critical path. 

t4 [was created](https://t4db.github.io/t4/why/) as a Kubernetes storage in the first place, as an attempt to simplify maintenance and introduce some new features and use-cases like branches and point-in-time restore.

pkg/drivers/t4/t4.go
```diff
@@ -0,0 +1,300 @@
...
+	"net/url"
+	"os"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+	awscredentials "github.com/aws/aws-sdk-go-v2/credentials"
+	awss3 "github.com/aws/aws-sdk-go-v2/service/s3"
```
brandond: For the K3s ecosystem we prefer minio-go; aws-sdk-go has in the past been responsible for significant bloat in the K3s binary size, and we would not accept anything that pulls in a new S3 and AWS credentials backend.

I realize it'd probably take some changes on the T4 side, but would it be possible to use `github.com/minio/minio-go/v7` instead?

See: 
https://github.com/k3s-io/k3s/blob/c6dd4c8f74a18c322cc61256e7ff178ac97ee026/pkg/etcd/s3/s3.go#L28-L29

pkg/drivers/t4/t4.go
```diff
@@ -0,0 +1,300 @@
...
+		opts = append(opts, awsconfig.WithEndpointResolverWithOptions(
+			aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
+				return aws.Endpoint{URL: endpoint}, nil
+			}),
+		))
+	}
+
+	accessKeyID := os.Getenv("T4_S3_ACCESS_KEY_ID")
+	secretAccessKey := os.Getenv("T4_S3_SECRET_ACCESS_KEY")
+	profile := os.Getenv("T4_S3_PROFILE")
```
brandond: I am curious why you'd prefix these with `T4_` instead of just using the normal env vars used to configure the AWS SDK?

Also, any configuration (from env vars or otherwise) should go in `pkg/app/` and be passed in via `endpoint.Config`: https://github.com/k3s-io/kine/blob/f29a4b3babbe1934ad9201b541511eef4f53f2a8/pkg/endpoint/endpoint.go#L38-L56

This allows us to consistently pass configuration to kine when running embedded, instead of having to wrangle env vars that it reads internally without direct control.

I believe you should also support `AWS_SESSION_TOKEN`? You can reference how we support S3 config for snapshots when using etcd: https://github.com/k3s-io/k3s/blob/c6dd4c8f74a18c322cc61256e7ff178ac97ee026/pkg/cli/cmds/server.go#L445-L462

pkg/drivers/t4/t4.go
```diff
@@ -0,0 +1,300 @@
...
+		return false, nil, fmt.Errorf("t4 driver: parse DSN: %w", err)
+	}
+
+	node, err := t4.Open(*nodeCfg)
+	if err != nil {
+		return false, nil, fmt.Errorf("t4 driver: open node: %w", err)
+	}
+
+	// leaderElect=false: t4 handles its own leader election via S3.
+	return false, &backend{node: node}, nil
```
brandond: That is not what the leaderElect retval means. When Kine is running embedded in k3s, the leader-elect bool tells K3s whether or not it should enable leader-election for Kubernetes components. Backends like sqlite that support only a single server can safely disable leader-election as there will only ever be a single server node. Backends with shared storage that support multi-server clusters must set this to true.

pkg/drivers/t4/t4.go
```diff
@@ -0,0 +1,300 @@
...
+func New(ctx context.Context, _ *sync.WaitGroup, cfg *drivers.Config) (leaderElect bool, b kserver.Backend, err error) {
+	nodeCfg, err := parseConfig(ctx, cfg.DataSourceName)
+	if err != nil {
+		return false, nil, fmt.Errorf("t4 driver: parse DSN: %w", err)
+	}
+
+	node, err := t4.Open(*nodeCfg)
+	if err != nil {
+		return false, nil, fmt.Errorf("t4 driver: open node: %w", err)
+	}
```
brandond, line 124: Is there a version of this that takes Context and WaitGroup for clean shutdown and contextual logging? Is there anything that T4 should do when stopping, such as explicitly flush writes and transfer leadership to another node? See the clean shutdown that other backends do, for example:
https://github.com/k3s-io/kine/blob/f29a4b3babbe1934ad9201b541511eef4f53f2a8/pkg/drivers/generic/generic.go#L201-L209

pkg/drivers/t4/t4.go
```diff
@@ -0,0 +1,300 @@
...
+	"google.golang.org/grpc/credentials"
+)
+
+func init() {
+	drivers.Register("t4", New)
+}
+
+// New is the kine driver constructor for the "t4" scheme.
+func New(ctx context.Context, _ *sync.WaitGroup, cfg *drivers.Config) (leaderElect bool, b kserver.Backend, err error) {
+	nodeCfg, err := parseConfig(ctx, cfg.DataSourceName)
```
brandond: The context passed here appears to be used for the S3 store, but not for the GRPC service - is that correct? Where does that context come from? How do you gracefully stop that?

brandond (commented):
Thanks, despite my initial concerns about S3 this does look interesting. I have some questions and a few things that we would need changed to accept this.

pkg/drivers/t4/t4.go
makhov: Yeah, makes sense. t4 currently accepts an `object.Store` interface and ships an aws-sdk-go-v2 implementation, but the interface itself is just a small set of S3 ops (Get/Put/Head/Delete/List). So basically, I can:
- Implement the Store interface in the driver using minio-go and not import t4's S3 package at all
- Switch t4 implementation to minio-go. 

I'm leaning towards the second option.

pkg/drivers/t4/t4.go
makhov: I decided to use `T4_S3_*` prefix to make it explicit and to avoid using user's aws creds by accident. Setting `T4_S3_PROFILE` (e.g. `T4_S3_PROFILE=default`) makes t4 to use the standard AWS chain (`AWS_*` env vars -> `~/.aws/credentials` -> IAM/IMDS). 
But I don't have a strong opinion on that

And I will add `AWS_SESSION_TOKEN` support, thanks!

pkg/drivers/t4/t4.go
makhov: Thank you, I got it wrong. Fixed!

pkg/drivers/t4/t4.go
makhov: Thanks, added!

pkg/drivers/t4/t4.go
makhov: Yes, the ctx from `New()` was only threaded into the AWS SDK config load — t4 itself takes a non-context `Open(Config)` and runs its own internal goroutines under an internal context. 

The fix is the same as in the comment above. Cancelling the kine ctx calls `Node.Close`, which drains
in-flight writes, flushes the WAL to S3, stamps the election lock with the final committed revision, then signals followers to take over (when this node is leader) or tells the leader the disconnect is intentional (when this node is a follower)

pkg/drivers/t4/t4.go
makhov: Switched t4 to minio-go, thanks!

pkg/drivers/t4/t4.go
makhov: Added `AWS_SESSION_TOKEN` support and moved settings to `endpoint.Config`. I kept it under `KINE_DATASTORE_T4_S3_*` env vars for now with an option to enable default AWS creds chain by setting `KINE_DATASTORE_T4_S3_PROFILE`, but I can change it if you prefer another approach.

pkg/drivers/t4/t4.go
brandond: That will need to be wired up to a waitgroup that explicitly waits for `node.Close()` to return in order to actually work. Otherwise the whole process just exits when the context is cancelled. See the bit above.

pkg/app/app.go
```diff
@@ -198,6 +198,36 @@ func New() *cli.App {
...
+			Usage:       "AWS profile name to engage the ambient AWS credentials chain for the t4 driver (env vars, ~/.aws/credentials, IMDS). Ignored when --datastore-t4-s3-access-key-id is set.",
+			Destination: &config.T4Config.Profile,
+			EnvVars:     []string{"KINE_DATASTORE_T4_S3_PROFILE"},
+		},
+		&cli.StringFlag{
+			Name:        "datastore-t4-s3-ca-bundle",
+			Usage:       "Path to a PEM CA bundle used to trust HTTPS S3 endpoints for the t4 driver. Useful for MinIO behind a private CA.",
+			Destination: &config.T4Config.CABundle,
+			EnvVars:     []string{"KINE_DATASTORE_T4_S3_CA_BUNDLE"},
+		},
```
brandond, line 266: Lets just use the normal AWS env var names? I'm not sure what you mean by `avoid using user's aws creds by accident` - this is meant to be run as a system service with intentional configuration, it's not some random CLI tool that someone is going to run and be surprised when it goes poking at their S3 buckets

I would also like to see a bunch of the T4 DSN parameters moved into configuration; DSNs should only configure the **client** connection to a backend datastore. Kine is the server and has first-class configuration.

* S3 config (region, endpoint, bucket name, prefix) should all have config fields
* peer TLS config should function like existing server/client tls flags
* peer-listen and peer-port should be combined and function similar to the existing bind-address and metrics-bind-address flags
* advertise-peer should have a config field

pkg/drivers/t4/t4.go
```diff
@@ -0,0 +1,273 @@
...
+func init() {
+	drivers.Register("t4", New)
+}
+
+// New is the kine driver constructor for the "t4" scheme.
+func New(ctx context.Context, wg *sync.WaitGroup, cfg *drivers.Config) (leaderElect bool, b kserver.Backend, err error) {
+	nodeCfg, err := parseConfig(ctx, cfg.DataSourceName, cfg.T4Config)
+	if err != nil {
+		return false, nil, fmt.Errorf("t4 driver: parse DSN: %w", err)
+	}
```
brandond: Do you need to use a new context for this, so that it's not using the already-cancelled context to do shutdown stuff? You can then cancel the independent context in the shutdown goroutine after calling node.Close().

brandond (changes requested):

scripts/test-run-t4-s3
```diff
@@ -0,0 +1,55 @@
...
+# database-pre-hook.
+database-pre-hook() {
+    echo "MINIO_ROOT_USER=minioadmin" >> $TEST_DIR/databases/$1/metadata/env
+}
+export -f database-pre-hook
+
+LABEL=t4-s3 \
+DB_IMAGE=quay.io/minio/minio:latest \
+DB_PASSWORD_ENV=MINIO_ROOT_PASSWORD \
+DB_ARGS="server /data --address :9000 --console-address :9001" \
```
brandond: We prefer s3mock for this, given recent issues with OSS versions of minio. See:
https://github.com/k3s-io/k3s/blob/c6dd4c8f74a18c322cc61256e7ff178ac97ee026/tests/e2e/s3/Vagrantfile#L32

brandond (changes requested):

scripts/test-run-t4-s3
makhov: Switched, thanks!

pkg/drivers/t4/t4.go
makhov: Done

pkg/app/app.go
makhov: Got it. I moved almost everything to the config. What I think should stay in the DSN:
- Bucket + prefix (`t4://bucket/prefix`): it's basically plays same role as mysql's `mysql://.../dbname`. Making user write `--endpoint t4:// --datastore-t4-s3-bucket my-bucket` feels a bit awkward.
- Datastore tuning (`data-dir`, `node-id`, `service-name`, `checkpoint-interval`, `segment-max-age`, `follower-max-retries`): again, same role as mysql's `?parseTime=...` or postgres's `?statement_cache_mode=...` or other params.

Also, I switched to usual `AWS_*` env vars.

One question: I currently use a t4-specific flags/vars for optional peer tls (`--datastore-t4-peer-cert-file` / `--datastore-t4-peer-key-file` / `--datastore-t4-peer-ca-file` /`--datastore-t4-peer-skip-verify`), but I can use existing flags: `--ca-file` / `--cert-file` / `--key-file` / `--skip-verify`. Should I use them?

brandond: I think bucket and prefix should go in config. From a functional perspective the UX of the AWS/S3 config should be similar to configuring etcd S3 snapshots.

data-dir should come from the path portion of the t4 dsn, for parity with the sqlite driver: `t4:///path/to/absolute/data-dir?node-id=foo` or `t4://path/to/relative/data-dir?node-id=foo`

I think reusing the client TLS args for T4 TLS config makes sense to me - since T4 nodes are basically "clients" to their peers.

For peer discovery -

pkg/app/app.go
```diff
@@ -198,6 +198,85 @@ func New() *cli.App {
...
+			Usage:       "AWS profile name to engage the ambient AWS credentials chain for the t4 driver (env vars, ~/.aws/credentials, IMDS). Ignored when --datastore-t4-s3-access-key-id is set.",
+			Destination: &config.T4Config.S3Profile,
+			EnvVars:     []string{"AWS_PROFILE"},
+		},
+		&cli.StringFlag{
+			Name:        "datastore-t4-s3-ca-bundle",
+			Usage:       "Path to a PEM CA bundle used to trust HTTPS S3 endpoints for the t4 driver. Useful for MinIO behind a private CA.",
+			Destination: &config.T4Config.S3CABundle,
+			EnvVars:     []string{"AWS_CA_BUNDLE"},
+		},
```
brandond, line 266: Also, take datastore-t4/DATASTORE_T4 out of the flag names and env vars. If we eventually have something else that needs S3 or peer listener config they can share these with T4. You can leave the mention of T4 in the descriptions, as these can be changed without making any breaking changes.

pkg/drivers/t4/t4.go
```diff
@@ -0,0 +1,287 @@
...
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		<-ctx.Done()
+		logrus.Infof("Closing t4 node...")
+		if err := node.Close(); err != nil {
+			logrus.Errorf("Failed to close t4 node: %v", err)
+		}
+	}()
```
brandond: Like this:

```suggestion
	nodeCtx, cancel := context.WithCancel(context.Background())
	nodeCfg, err := parseConfig(nodeCtx, cfg.DataSourceName, cfg.T4Config)
	if err != nil {
		return false, nil, fmt.Errorf("t4 driver: parse DSN: %w", err)
	}

	node, err := t4.Open(*nodeCfg)
	if err != nil {
		return false, nil, fmt.Errorf("t4 driver: open node: %w", err)
	}

	wg.Add(1)
	go func() {
		defer wg.Done()
		defer cancel()
		<-ctx.Done()
		logrus.Infof("Closing t4 node...")
		if err := node.Close(); err != nil {
			logrus.Errorf("Failed to close t4 node: %v", err)
		}
	}()
```

brandond (changes requested):

pkg/drivers/t4/t4.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	q := u.Query()
+
+	cfg := &t4.Config{
+		DataDir: q.Get("data-dir"),
+	}
+	if cfg.DataDir == "" {
+		cfg.DataDir = "/var/lib/t4"
+	}
```
brandond: For compat with sqlite - default to `db` dir in current working dir

```suggestion
	if cfg.DataDir == "" {
		cfg.DataDir = "db"
	}
```

pkg/app/app.go
makhov: Fair point, and I'm happy to flip it if you'd rather, but I see bucket+prefix as the address of the data, like the DB URI inb `mysql://user:pass@host:port/dbname` or the file path in `sqlite:///var/lib/state.db`. It's how you'd identify a specific deployment or point a new node at an existing cluster. Moving it out of the endpoint means `--endpoint t4://` no longer uniquely identifies the datastore, which feels off vs the other drivers.

The local data-dir on the other hand is more like a cache - a node that loses its disk rebuilds it from S3 on next start, no operator action.

But honestly, I'm not attached, so your final call, and I'll implement whichever way you prefer.

brandond: I think I'd prefer to keep the S3 stuff together in CLI config, and the data-dir path in the DSN, like sqlite does.

As you said, the live data that is replicated and queried actually lives on disk at the data-dir, same as with sqlite. It is checkpointed to and restored from S3, same as the K3s etcd snapshots. The parity is pretty clear.

If we wire this all up right, then T4 could be essentially a drop-in replacement for multi-node etcd + s3 snapshots, using the same configuration flags - just replace `--cluster-init` or `--server` with `--datastore-endpoint=t4://` and the correct S3 config.

brandond: That reminds me - does T4 have any way to validate a node that wants to join the cluster? With etcd, the joining member needs to prove that it has the correct server join token in order to obtain bootstrap data that contains the certificates necessary to connect to etcd, join the cluster, and start replicating data. Is there any similar concept with T4?

makhov: > I think I'd prefer to keep the S3 stuff together in CLI config

Done.

Regarding validating a node joining the cluster - t4 currently relies on AWS creds and peer tls

pkg/app/app.go
makhov: Done

pkg/drivers/t4/t4.go
makhov: Got it. Done

pkg/drivers/t4/t4.go
makhov: Done

pkg/drivers/config.go
```diff
@@ -21,4 +21,59 @@ type Config struct {
...
+	// node's peer server. Defaults to PeerBindAddress.
+	PeerAdvertiseAddress string
+
+	// S3Bucket is the S3 bucket name. Empty means local-only mode
+	// (no S3 durability).
+	S3Bucket string
+
+	// S3Folder is an optional key prefix inside the bucket (no trailing
+	// slash needed). Equivalent to k3s's --etcd-s3-folder.
+	S3Folder string
```
brandond: Final nit: split this into PeerConfig and S3Config.
```suggestion
type PeerConfig struct {
	// BindAddress is the gRPC listen address (host:port) for t4's
	// WAL-streaming server, e.g. "0.0.0.0:3380". Empty means single-node
	// mode (no peer server).
	BindAddress string

	// AdvertiseAddress is the address other T4 nodes use to reach this
	// node's peer server. Defaults to BindAddress.
	AdvertiseAddress string
}

type S3Config struct {
	// Bucket is the S3 bucket name. Empty means local-only mode
	// (no S3 durability).
	Bucket string

	// Folder is an optional key prefix inside the bucket (no trailing
	// slash needed). Equivalent to k3s's --etcd-s3-folder.
	Folder string
```

brandond: Question about AdvertiseAddress - how does it detect what IP address to advertise if you bind to `0.0.0.0` or `::`?

brandond (changes requested):
final nits

pkg/drivers/config.go
makhov: If `service-name` is set, it sets `<hostname>.<service-name>:<port>`, but otherwise It doesn't detect anything, just falls back to `BindAddress`.

I can make it required if we bind to `0.0.0.0/::`, or just make it required always.

makhov: Made it required

pkg/drivers/config.go
```diff
@@ -21,4 +21,62 @@ type Config struct {
...
+
+	// S3Profile engages the ambient AWS credentials chain — env vars
+	// (AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY / AWS_SESSION_TOKEN), then
+	// the named profile in ~/.aws/credentials, then EC2/EKS IMDS.
+	S3Profile string
+
+	// S3CABundle is an optional PEM-encoded CA-certificate bundle path used
+	// to trust HTTPS S3 endpoints (MinIO behind a private CA, etc.). Empty
+	// means the system trust store.
+	S3CABundle string
```
brandond: sorry - I figured it was implicit, but the rest of these fields should also have the `S3` prefix removed since they are in a struct that is already called S3.

brandond:
Have you not tested this with the core Kubernetes apiserver tests yet? You can run `(source ./scripts/test-helpers; build-apiserver-tests)` to build these locally.

makhov:
Thanks for suggestion! I found and fixed some real gaps in t4 driver behavior. Now the apiserver tests pass locally.

scripts/test-run-t4-s3
```diff
@@ -0,0 +1,60 @@
...
+    local ip=$(cat $TEST_DIR/databases/*/metadata/ip)
+    local bucket="kine-t4-test"
+
+    DB_CONNECTION_TEST="
+        docker run --rm
+        --name connection-test
+        --network container:$(cat $TEST_DIR/databases/*/metadata/name)
+        --entrypoint sh
+        curlimages/curl:8.10.1
+        -c \"curl --silent --fail http://localhost:9090/${bucket}\"" \
```
brandond: try this>

```suggestion
    local bucket="kine-t4-test"

    DB_CONNECTION_TEST="
        docker run --rm
        --name connection-test
        --network container:$(cat $TEST_DIR/databases/*/metadata/name)
        --entrypoint sh
        curlimages/curl:8.10.1
        -c 'curl --silent --fail http://localhost:9090/'${bucket}" \
```

makhov: Thanks, I think it's fixed now

brandond (approved):

brandond:
Looks good! Would you mind squashing down to one commit, now that everything is green?

makhov:
Thanks! I squashed it

pkg/drivers/t4/README.md
```diff
@@ -0,0 +1,228 @@
...
+headless services, issuing one wildcard cert for `*.<service-name>` is the
+simplest path.
+
+`--skip-verify` disables peer-cert verification on the client side and
+downgrades the server to `tls.RequestClientCert` — useful for short-lived
+debug clusters, but never enable it in production.
+
+## S3 Credentials
+
+S3 credentials live on `drivers.Config.T4Config` (CLI flag / env / programmatic
```
galal-hussein:
```suggestion
S3 credentials live on `drivers.Config.S3Config` (CLI flag / env / programmatic
```

pkg/drivers/t4/backend.go
```diff
@@ -0,0 +1,286 @@
...
+	case errors.Is(err, t4.ErrFutureRevision):
+		return kserver.ErrFutureRev
+	default:
+		return err
+	}
+}
+
+func (b *backend) Create(ctx context.Context, key string, value []byte, lease int64) (int64, error) {
+	rev, err := b.node.Create(ctx, key, value, lease)
+	if err == t4.ErrKeyExists {
```
galal-hussein: Is there a reason not using errors.Is like in translateReadErr()?

galal-hussein (approved):
couple of nits, but LGTM overall, thanks!

brandond:
👍🏻 on nits from @galal-hussein, if you want to address those and squash again then I think we are good to merge

vitorsavian (approved):

pkg/drivers/t4/README.md
makhov: Fixed, thanks!

pkg/drivers/t4/backend.go
makhov: Not really, switched it to errors.Is, thanks!

Issue #556, opened 2025-11-29
It's time to fill out the pure go sqlite driver for better portability


viscropst:
As title, current version of kine ignores sqlite without cgo. But sqlite has a great pure go implemention `modernc.org/sqlite` can handle almost time of kine use. And [k0s-project/k0s#6730](https://github.com/k0sproject/k0s/issues/6730) want to disable all cgo build cause I can use the etcd or other dbms cannot use sqlite for better memory usage and easier deployment, I usually use the kine for using the sqlite instead of etcd. So most time of kine usage will break.

PR #678, opened 2026-05-28
Add pure-go (-nocgo) sqlite variant

brandond:
This is a from-scratch rework of https://github.com/k3s-io/kine/pull/557 with the unnecessary changes left out.

* Uses sqlite from modernc.org/sqlite
* Adds -nocgo binaries and image tags
* Same CI coverage as the current cgo sqlite variant from github.com/mattn/go-sqlite3

Other than package names, the only differences between variants are:
* WAL fcntl callback interface (for litespeed)
* Format of pragmas in DSN query string
* Error code translation

Linked issues:
* https://github.com/k3s-io/kine/issues/556
* https://github.com/k0sproject/k0s/issues/6730

Note: the cgo variant is actually 1mb smaller:
```console
-rwxr-xr-x  1 brandond brandond 47263816 May 29 05:25 kine
-rwxr-xr-x  1 brandond brandond 48349368 May 29 05:22 kine-nocgo
```


dereknola (changes requested):
Can we duplicate the `sqlite_cgo_test.go` to also cover some of the pure-go sqlite implementation? Even if its a duplication of the unit test, we should have some coverage of these funtions.

.github/workflows/pr.yml
```diff
@@ -62,32 +67,35 @@ jobs:
...
       with:
-        name: images-${{ matrix.arch }}
+        name: images-${{ matrix.arch }}${{ env.SUFFIX }}
         path: ${{ runner.temp }}/kine*.tar
         if-no-files-found: error
         retention-days: 1
 
   docker:
-    name: Docker
+    name: Test ${{ matrix.test }} (ARCH=${{ matrix.arch }} CGO_ENABLED=1)
```
dereknola: Adding the CGO status makes reading these really difficult on the actions page. And it is redundant, because the name of the job itself already contains CGO information. Also these just don't have space on the graph, so get rid of the `ARCH=`, since it will be very clear what `arm64` and `amd64` by themselves are. 
```suggestion
    name: Test ${{ matrix.test }} (${{ matrix.arch }})
```

Image

dereknola (changes requested):

.github/workflows/pr.yml
brandond: That only helps with the graph view, in the sidebar the matrix name is not shown. Also, if you don't include the matrix elements in the name, it appends them comma-separated so you get `Build (amd64, 0)` which is not super helpful if you don't know what the 0 or 1 means.

dereknola (approved):

vitorsavian (approved):

PR #682, opened 2026-06-01
Don't cache linters

dereknola:
Both `golangci-lint-action` and the `raven-actions/actionlint` use caching by default. However, as these actions are only ever run on PRs, the cache can never get reused and is always invalidated by the next PR. This leads to the cache being filled with useless entries. 

image

Signed-off-by: Derek Nola <derek.nola@suse.com>

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -23,6 +23,7 @@ jobs:
       uses: golangci/golangci-lint-action@82606bf257cbaff209d206a39f5134f0cfbfd2ee # v9
       with:
         version: v2.7
+        skip-cache: true
 
     - name: Validate
       run: ./scripts/validate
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yml
```diff
@@ -23,6 +23,7 @@ jobs:
       uses: golangci/golangci-lint-action@82606bf257cbaff209d206a39f5134f0cfbfd2ee # v9
       with:
         version: v2.7
+        skip-cache: false
```
brandond: wait, do we want skip the cache, or not skip the cache?

dereknola: Ugn double negative.

brandond (approved):

vitorsavian (approved):

PR #692, opened 2026-06-15
Fix SQLite name affinity and LIKE optimization

### LINKED ISSUE #690: sqlite: name column declared INTEGER instead of TEXT; missing PRAGMA case_sensitive_like prevents LIKE optimization
## Summary

Two related issues in the SQLite driver (`pkg/drivers/sqlite/sqlite.go`):

1. The `kine` table `name` column is declared `INTEGER` instead of `TEXT`
2. `PRAGMA case_sensitive_like` is not set to `ON`, preventing the LIKE optimization from applying

## 1. `name` column should be `TEXT`, not `INTEGER`

**Location**: `pkg/drivers/sqlite/sqlite.go:27`

```go
name INTEGER,   // should be TEXT
```

The other two drivers use proper string types:
- **PostgreSQL**: `name TEXT COLLATE "C"`
- **MySQL**: `name VARCHAR(630) CHARACTER SET ascii`

SQLite uses manifest typing so string values store and retrieve correctly regardless — but the INTEGER affinity is incorrect and could cause edge-case query plan issues if key names ever contained well-formed numeric literals.

**Empirically**: No practical impact on current kine workloads because all key names start with `/` (non-numeric), so SQLite skips the numeric affinity conversion. Query plans and wall-clock timing are identical between `INTEGER` and `TEXT` on a 66MB / 38K-row production database. Still worth fixing for consistency and future-proofing.

## 2. Missing `PRAGMA case_sensitive_like = ON`

**Location**: `pkg/drivers/sqlite/sqlite.go:setup()` (~line 130)

SQLite's [LIKE optimization](https://sqlite.org/optoverview.html#like) converts prefix LIKE patterns into range scans — but only when the indexed column uses BINARY collation AND `case_sensitive_like` is ON (default is OFF).

Without this PRAGMA, every ListCurrent prefix query does a **full index scan**:

```
SCAN mkv USING COVERING INDEX kine_name_index
```

With `PRAGMA case_sensitive_like = ON`, it uses a **range scan**:

```
SEARCH mkv USING COVERING INDEX kine_name_index (name>? AND name<?)
```

**Impact**: On larger databases (multi-GB, millions of rows, as reported in issues #213, #5033, #607), full index scans for every prefix list operation contribute to slow reads → compaction timeouts → unbounded database growth.

**Safety**: Kubernetes resource paths are lowercase ASCII — case-sensitive LIKE is correct behavior. Verified with all critical kine query patterns (ListCurrent, After poll, compact_rev_key lookup, compaction DELETE filter) on a 66MB database — all produce identical results with the PRAGMA enabled. GLOB (which is always case-sensitive) also produces identical results.

### Fix

In `pkg/drivers/sqlite/sqlite.go`, in the `setup()` function:

```diff
        if noAutoCheckpoint {
                logrus.Infof("WAL auto-checkpoint is disabled")
                schema = append(schema, `PRAGMA wal_autocheckpoint = 0`)
        }

+       if _, err := db.Exec(`PRAGMA case_sensitive_like = ON`); err != nil {
+               return err
+       }
+
        for _, stmt := range schema {
```

Optionally also fix the column type in the DDL (requires migration for existing databases):

```diff
-               name INTEGER,
+               name TEXT,
```

### EXPLAIN QUERY PLAN comparison

**Before** (default `case_sensitive_like = OFF`):
```
CO-ROUTINE maxkv
  `--SCAN mkv USING COVERING INDEX kine_name_index
```

**After** (`PRAGMA case_sensitive_like = ON`):
```
CO-ROUTINE maxkv
  `--SEARCH mkv USING COVERING INDEX kine_name_index (name>? AND name<?)
```

### Other approaches considered

- **GLOB**: Also gets the range scan with BINARY indexes and no PRAGMA needed. But GLOB doesn't support ESCAPE clause (kine uses `ESCAPE '^'`) and uses different wildcards (`*` vs `%`).
- **NOCASE indexes**: Not appropriate — would break case-sensitive key matching for Kubernetes resource paths.

### References

- [SQLite LIKE optimization docs](https://sqlite.org/optoverview.html#like)
- [SQLite type affinity docs](https://sqlite.org/datatype3.html)
- [SQLite Quirks - INTEGER + text storage](https://sqlite.org/quirks.html)
- [StackOverflow: 100x slowdown from INTEGER affinity on text column](https://stackoverflow.com/questions/22060197)
- [kine PR #323: PG collation fix for LIKE](https://github.com/k3s-io/kine/pull/323)
- [kine issue #113: Superfluous index definition](https://github.com/k3s-io/kine/issues/113)

wyf027:
Closes k3s-io/kine#690.

## Summary
- Declare the SQLite `kine.name` column as `TEXT` for newly created databases.
- Enable `case_sensitive_like` for SQLite connections so prefix LIKE queries can use range scans.
- Translate the same setting for the modernc/sqlite no-cgo driver path.
- Add SQLite setup tests for the column type and LIKE behavior.

## Verification
- `gofmt -w pkg/drivers/sqlite/sqlite.go pkg/drivers/sqlite/sqlite_nocgo.go pkg/drivers/sqlite/sqlite_test.go`
- `go test ./pkg/drivers/sqlite`
- `CGO_ENABLED=0 go test ./pkg/drivers/sqlite`
- `git diff --check`

### DIFF
--- a/pkg/drivers/generic/generic.go
+++ b/pkg/drivers/generic/generic.go
@@ -38,9 +38,10 @@ var (
 		WHERE (deleted = 0 OR ?)
 		ORDER BY name ASC
 		`
-	CurrentRevSQL = `SELECT MAX(id) AS current_rev FROM kine`
-	CompactRevSQL = `SELECT MAX(prev_revision) AS compact_rev FROM kine WHERE name = 'compact_rev_key'`
-	FiltersRevSQL = `SELECT MAX(id) AS id FROM kine WHERE name LIKE ? ESCAPE '^' %s GROUP BY name`
+	CurrentRevSQL      = `SELECT MAX(id) AS current_rev FROM kine`
+	CompactRevSQL      = `SELECT MAX(prev_revision) AS compact_rev FROM kine WHERE name = 'compact_rev_key'`
+	FiltersRevSQL      = `SELECT MAX(id) AS id FROM kine WHERE name LIKE ? ESCAPE '^' %s GROUP BY name`
+	FiltersRevRangeSQL = `SELECT MAX(id) AS id FROM kine WHERE name >= ? AND name < ? %s GROUP BY name`
 
 	listSQL    = fmt.Sprintf(ListFmt, Columns, CurrentRevSQL, CompactRevSQL, FiltersRevSQL)
 	listValSQL = fmt.Sprintf(ListFmt, WithVal, CurrentRevSQL, CompactRevSQL, FiltersRevSQL)
@@ -50,6 +51,7 @@ type ErrRetry func(error) bool
 type TranslateErr func(error) error
 type ErrCode func(error) string
 type SubstituteFunc func(string) string
+type PrefixRangeFunc func(string) (string, string, bool)
 
 type ConnectionPoolConfig struct {
 	MaxIdle     int           // zero means defaultMaxIdleConns; negative means 0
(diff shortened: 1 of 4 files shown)


brandond (changes requested):
Please see the upstream docs:

https://www.sqlite.org/pragma.html#pragma_case_sensitive_like
> This pragma is deprecated and exists for backwards compatibility only. New applications should avoid using this pragma. Older applications should discontinue use of this pragma at the earliest opportunity.

Given that warning and deprecation, this does NOT seem to be the correct way to implement this enhancement.

brandond:
From the upstream docs:

https://sqlite.org/pragma.html#pragma_case_sensitive_like
> This pragma uses [sqlite3_create_function()](https://www.sqlite.org/c3ref/create_function.html) to overload the LIKE and GLOB functions, which may override previous implementations of LIKE and GLOB registered by the application.

https://sqlite.org/optoverview.html#the_like_optimization
> * The built-in functions used to implement LIKE and GLOB **must not** have been overloaded using the [sqlite3_create_function()](https://sqlite.org/c3ref/create_function.html) API.
For the GLOB operator, the column must be indexed using the built-in BINARY collating sequence.
> * For the LIKE operator, if [case_sensitive_like](https://sqlite.org/pragma.html#pragma_case_sensitive_like) mode is enabled then the column must be indexed using the built-in BINARY collating sequence, or *if [case_sensitive_like](https://sqlite.org/pragma.html#pragma_case_sensitive_like) mode is disabled then the column must be indexed using the built-in NOCASE collating sequence.*

https://sqlite.org/optoverview.html#the_like_optimization
> Let y be the smallest string that is the same length as /x/ but which compares greater than x. For example, if x is 'hello' then y would be 'hellp'. The LIKE and GLOB optimizations consist of adding two virtual terms like this:
> `column >= x AND column < y`

I think we're already doing something like this for ranges; rather than faffing about with pragmas and LIKE we could probably just convert everything to use a between on the name. Logic already exists do to this for range requests.

wyf027:
Updated this to remove the deprecated case_sensitive_like PRAGMA and use explicit name range predicates for SQLite prefix scans instead. I also kept the name column as TEXT and added tests for the prefix range conversion and range query behavior.

brandond:
I have a more comprehensive version of this that does away with prefix/startKey entirely, and just wholesale converts to BETWEEN:
* https://github.com/k3s-io/kine/pull/693

wyf027:
Closing this in favor of #693, which is the maintainer-owned, more comprehensive BETWEEN-based replacement for this change. Thanks for reviewing and pointing the implementation in the right direction.

PR #698, opened 2026-06-23
Enable sqlite internal statement cache

brandond:
x-ref: https://github.com/mattn/go-sqlite3/pull/1387

> This change adds an opt-in connection-local statement cache to go-sqlite3, enabled via the DSN parameter _stmt_cache_size=N.
> 
> The motivation is that, in this benchmark, the main gap in mattn/go-sqlite3 was not cgo alone but repeated prepare/finalize on hot QueryRow paths. Reusing prepared statements at the driver level closes most of that gap without requiring application code to switch to explicit db.Prepare(...).

### DIFF
--- a/pkg/drivers/sqlite/sqlite.go
+++ b/pkg/drivers/sqlite/sqlite.go
@@ -19,7 +19,7 @@ import (
 )
 
 var (
-	DefaultParams = "_journal_mode=WAL&_busy_timeout=30000&_synchronous=NORMAL&_txlock=immediate&cache=shared"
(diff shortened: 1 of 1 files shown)


manuelbuil (commented):
why 20?

brandond:
It's an LRU and doesn't consume much resources... and we don't have very many distinct queries... so I some what arbitrarily picked 20 to match the default connection pool size.

https://github.com/k3s-io/kine/blob/6edfa5feedf4e4c80c2550a652f6f59aea500694/pkg/drivers/generic/generic.go#L24

matttrach (approved):

Issue #720, opened 2026-07-28
NATS backend: btree replay times out at 60s on large KV_kine stream (25M sequences); no way to raise timeout or force compaction before readiness


christiangierschner:
**Summary**: k3s enters a permanent systemd restart loop. On every start, the kine NATS backend begins an initial btree replay over the full stream sequence range and is killed by an internal 60-second readiness timeout before replay completes. Because each restart replays from sequence 0, no progress is ever made.

**Environment**:

    k3s: v1.35.6+k3s1 (commit 87243446), Go 1.25.11
    Datastore backend: embedded kine → NATS JetStream
    NATS server: v2.10.7, 3-node cluster (cluster1)
    OS: Ubuntu 24.04
    Node role: server (control-plane), 3-server HA-style datastore

**Datastore endpoint**:
```
K3S_DATASTORE_ENDPOINT='nats://clusternode-1:4222,nats://clusternode-2:4222,nats://clusternode-3:4222'
```
(no query parameters configured)

**Fatal log (reproduced identically on every restart, including foreground k3s server --debug, so it is NOT systemd killing the process):**

```
INFO[0000] Starting k3s v1.35.6+k3s1 (87243446)
DEBU[0000] using config &nats.Config{clientURL:"nats://...:4222,...", revHistory:0xa, bucket:"kine", replicas:1, slowThreshold:500000000, noEmbed:false, dontListen:false, serverConfig:"", host:"...", port:4222, dataDir:""}
INFO[0000] connecting to nats://...
INFO[0000] using bucket: kine
INFO[0000] bucket initialized: kine
INFO[0000] compactInterval: 0s, compactMinRetain: 1000
INFO[0000] started expire watcher
INFO[0000] Waiting for btree replay to complete...
INFO[0000] nats: starting initial replay from 0 to 25367210
INFO[0060] Shutdown request received
INFO[0060] nats: stopping key value store
INFO[0060] Waiting up to 2s for graceful shutdown of Kine GRPC server...
FATA[0060] Error: preparing server: failed to bootstrap cluster data: creating storage endpoint: starting kine backend: failed to initialize btree: timeout waiting for btree to be ready
```

**JetStream stream state (nats stream info KV_kine):**
```
Messages:        307,279   (live)
Bytes:           1.3 GiB
First Sequence:  3
Last Sequence:   25,367,210
Deleted Messages: 25,059,929
Number of Subjects: 77,361
Replicas:        1
Max Per Subject: 10   (revHistory)
[...]

PR #722, opened 2026-07-29
Optimize nats replay decoding

bruth:
For #720, based on a scale testing benchmark, a core bottleneck was S2 reader allocations. This change reducing the default allocation size of the reader itself, but also introduces a sync.Pool to reuse existing buffers.

### DIFF
--- a/pkg/drivers/nats/codec.go
+++ b/pkg/drivers/nats/codec.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"sync"
 
 	"github.com/klauspost/compress/s2"
 	"github.com/nats-io/nats.go/jetstream"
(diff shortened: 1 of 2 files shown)


brandond:
* Can you put some text in the PR description? I believe this is for https://github.com/k3s-io/kine/issues/720 but some commentary on what this does would be good.

bruth:
@brandond I updated the description and rebased against master.

pkg/drivers/nats/codec.go
```diff
@@ -4,18 +4,27 @@ import (
...
 	"github.com/klauspost/compress/s2"
 	"github.com/nats-io/nats.go/jetstream"
 	"github.com/shengdoushi/base58"
 )
 
 const (
-	noRootPrefix = "meta"
+	noRootPrefix              = "meta"
+	valueDecoderAllocBlock    = 1024
+	valueDecoderPoolMaxRetain = 65_536
```
brandond: nit: no separators in integer literals
```suggestion
	valueDecoderPoolMaxRetain = 65536
```

pkg/drivers/nats/codec_test.go
```diff
@@ -95,3 +98,99 @@ func TestKeyEncodeRange(t *testing.T) {
 		}
 	}
 }
+
+func TestValueCodecRoundTrip(t *testing.T) {
+	for _, size := range []int{0, 1, 1_024, 4_096, 65_536, 1_048_577} {
```
brandond: same nit - no blank separators in integer literals

brandond (changes requested):

brandond (approved):

Issue #246, opened 2023-11-17
feat: IAM database authentication


a7i:
Support IAM database authentication: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html

kine only supports username/password or cert-based auth which makes it a bit challenging for organizations with compliance to use short-lived credentials. Supporting IAM database auth eases the rotation and management of credentials.

brandond:
Can you suggest any golang mysql or postgres drivers that support this? It looks fairly non-trivial to enable, as credentials are currently part of the datastore dsn and we leave their extraction up to the driver.

a7i:
Hey @brandond I see, I did not realize that kine just relies on the drivers (it makes sense). So I would need to make the change in this driver: https://github.com/go-sql-driver/mysql

brandond:
and `github.com/jackc/pgx` for postgres

a7i:
Thanks @brandond 🎉 , I'll go ahead and close this Issue then

PR #726, opened 2026-07-31
pgsql: add AWS RDS IAM auth support

ianb-mp:
When no password is supplied in the Postgres datastore DSN, attempt to source credentials from the AWS default credential provider chain and authenticate via RDS IAM.

I tested this from an ec2 instance against an RDS Postgres instance: with kine running in the backgound, I setup `etcdctl` to write a new value every 10s in a loop for ~2hrs. No errors were encountered.

```
$ ./kine --metrics-bind-address ":8081" --endpoint "postgres://iam_user_ianb@esp-k0s-postgres-db.xxxxxxx.ap-southeast-2.rds.amazonaws.com:5432/kine_ianb?sslmode=verify-full&sslrootcert=rds-global-bundle.pem"
INFO[2026-07-31T04:22:46.398415028Z] No password supplied in datastore DSN; using AWS default credential provider to generate RDS IAM authentication tokens for user "iam_user_ianb"
INFO[2026-07-31T04:22:46.4860671Z] Configuring pgx database connection pooling: maxIdle=20, maxOpen=0, maxLifetime=0s, maxIdleTime=2m0s
INFO[2026-07-31T04:22:46.486749492Z] Configuring database table schema and indexes, this may take a moment...
INFO[2026-07-31T04:22:46.49085544Z] Database tables and indexes are up to date
INFO[2026-07-31T04:22:46.523434497Z] Kine available at http://127.0.0.1:2379
INFO[2026-07-31T04:22:46.523631494Z] metrics server is starting to listen at :8081
INFO[2026-07-31T04:22:46.523825468Z] starting metrics server path /metrics
```

A similar approach should work for MySQL however I don't have a server to test against, so this is Postgres only for now.

Relates to https://github.com/k3s-io/kine/issues/246

Code was written with assistance from Claude Opus 4.8


brandond:
Thanks for the PR!

See the note at https://github.com/k3s-io/kine/pull/669#discussion_r3255300477 about use of aws-sdk-go.

For the K3s ecosystem we prefer minio-go; aws-sdk-go has in the past been responsible for significant bloat in the K3s binary size, and we would not accept anything that pulls in a new AWS credentials backend. Please use github.com/minio/minio-go/v7 AWS credential helpers instead.

example: https://github.com/t4db/t4/blob/3ddb606a09a6fe720c06b419845f3c4598224a22/pkg/object/s3.go#L119-L144

I'm not sure if this library has all the bits for RDS auth in particular since it is more focused on S3, but we definitely don't want aws-sdk-go.

ianb-mp:
ok, I'll take another look. I assume this repo's AI policy is the same as k3s? (I'll may lean on Claude for the missing bits) https://github.com/k3s-io/k3s/blob/main/CONTRIBUTING.md#ai-guidance

brandond:
Yeah, same AI policy. No slop, no excessive verbosity or self-indulgent tests, make sure it does what it says on the tin.

ianb-mp:
Pushed https://github.com/k3s-io/kine/pull/726/commits/41690f3e0e6ef74f17a4d05dd5bca594eaeed519 which migrates code to minio-go. I ran the same test as described in the PR body.

pkg/drivers/pgsql/aws.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// awsIAMConnOptions returns the pgx stdlib connection options required to
+// authenticate to Postgres via RDS IAM when no password is supplied in the
+// datastore DSN.
+//
+// Returns nil if no AWS region is configured, i.e. we are not in an AWS
+// environment and RDS IAM was not intended.
+func awsIAMConnOptions(_ context.Context, config *pgx.ConnConfig) ([]stdlib.OptionOpenDB, error) {
```
brandond, line 22: Is all of this postgres-specific, or does AWS also support this for other RDS database types? If AWS supports this for all engines, we should support it for all of them.

pkg/aws/imds.go
```diff
@@ -0,0 +1,102 @@
...
+	imdsTimeout = 5 * time.Second
+
+	// imdsDefaultEndpoint is the link-local base URL of the EC2 instance
+	// metadata service (IMDS).
+	imdsDefaultEndpoint = "http://169.254.169.254" //nolint:revive // HTTPS is not supported by IMDS.
+	imdsTokenPath       = "/latest/api/token"
+	imdsRegionPath      = "/latest/meta-data/placement/region"
+	imdsTokenTTLHeader  = "X-aws-ec2-metadata-token-ttl-seconds"
+	imdsTokenHeader     = "X-aws-ec2-metadata-token"
+	imdsTokenTTL        = "21600"
```
brandond: There's not a library for any of this? I am pretty sure that the minio-go credentials helpers support loading from ambient creds (IMDS/instance role), why do we need all this here?

pkg/aws/aws.go
```diff
@@ -0,0 +1,188 @@
...
+	rdsIAMTokenExpirySeconds = 900 // 15 minutes
+	// sigV4Algorithm is the SigV4 signing algorithm identifier.
+	sigV4Algorithm = "AWS4-HMAC-SHA256"
+	// iso8601BasicFormat and shortDateFormat are the timestamp layouts SigV4
+	// expects for the X-Amz-Date query parameter and the credential scope.
+	iso8601BasicFormat = "20060102T150405Z"
+	shortDateFormat    = "20060102"
+	// emptyPayloadHash is the SHA-256 of an empty body, which RDS expects for
+	// the bodyless presigned "connect" request.
+	emptyPayloadHash = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
```
brandond, line 37: This doesn't look RDS specific, isn't this just the standard AWS SigV4 stuff? Please confirm that there's not anything in minio-go or elsewhere that we can use without having to maintain this here. I REALLY don't want to own a bunch of AWS library code.

brandond (changes requested):

pkg/drivers/pgsql/aws.go
ianb-mp: AWS supports MySQL/MariaDB also. I don't use that type so have nothing to test against. I was thinking that could be added later by someone who can test it.

pkg/aws/imds.go
ianb-mp: You're right, some of these constants are covered by minio-go [credentials](https://pkg.go.dev/github.com/minio/minio-go/v7@v7.1.0/pkg/credentials) package - I'll push an update.

pkg/aws/aws.go
ianb-mp: minio-go has a PresignV4() function but it's [hard coded to S3](https://github.com/minio/minio-go/blob/7e34713241bbd80c4cae66ad141d232c0623387b/pkg/signer/request-signature-v4.go#L249). It also doesn't export some relevant bits e.g. [signV4Algorithm](https://github.com/minio/minio-go/blob/e7b08014a7ca6bd0c3acf64774937529eaf4a682/constants.go#L62). I guess this is the necessary tradeoff for not using AWS SDK.

pkg/drivers/pgsql/aws.go
brandond: Please wire it up anyway. I'm not sure how we're going to test any of this locally. I generally prefer not to add things we can't test in CI, but for this perhaps we can make an exception.

PR #728, opened 2026-07-31
Fix delayed watch progress notifications

brandond:
Fixes issue where notifications waiting on watch to sync to revision would not wake until the next poll cycle - which could be up to 1 second if no inserts occur to wake the poller early.

Also replaces dedicated progress-request polling goroutine with on-demand goroutine created when new progress is requested. CAS on the notification bool ensures that multiple goroutines do not stack.

The combination of the two delays (100ms notify interval, plus up to 1sec delay for sync) were creating significant delays in DeleteCollection requests, which are used by the namespace controller to remove all resources from a namespace when it is deleted. DeleteCollection relies on progress notification requests to ensure that apiserver caches are synced with etcd before deleting things.

K3s issue:
* https://github.com/k3s-io/k3s/issues/14399

### DIFF
--- a/pkg/broadcaster/cond.go
+++ b/pkg/broadcaster/cond.go
@@ -1,0 +1,44 @@
+package broadcaster
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// Cond is a wrapper around sync.Cond that facilitates
+// waiting for a broadcaster to signal that is has seen
+// a specific revision.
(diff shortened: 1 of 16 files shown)


brandond:
This is going to take a major rewrite of the watch pipeline. See https://github.com/k3s-io/kine/issues/731

brandond:
Closing this, as it is going to require major refactoring to fix without introducing races.

PR #736, opened 2026-08-07
Surface rows.Err() so interrupted queries are not read as empty results

nic-6443:
`RowsToEvents` never checks `rows.Err()` after the `rows.Next()` loop, so a query that dies mid-flight — killed session, deadlock victim, cancelled statement, dropped connection — is indistinguishable from an empty result set.

That empty read propagates as "key does not exist": `logstructured.Delete` treats it as "already deleted" and returns success **without appending a tombstone**, so the client gets `Succeeded: true` while the key is still readable afterwards. Point Gets and prefix Lists can likewise come back spuriously empty, and the CAS create/update flows then run against a wrong base.

I hit this in production-like testing on the SQL Server backend and could reproduce it reliably: run kine over SQL Server while a chaos loop KILLs active sessions on the database every 40ms, with workers doing create → verify → delete cycles through the normal gRPC path. In a 90s run (~6k kills), 72 deletes reported success while the row provably stayed live in the kine table, plus 41 Gets that returned empty right after a successful create. With this patch the same run shows zero anomalies — the interrupted statements surface as errors instead, which clients can retry.

The added tests use a fake driver that fails from `driver.Rows.Next` before the first row and mid-stream; both cases return a nil error on master today.

### DIFF
--- a/pkg/logstructured/sqllog/sql.go
+++ b/pkg/logstructured/sqllog/sql.go
@@ -407,6 +407,16 @@ func RowsToEvents(rows *sql.Rows, val, prev bool) (int64, int64, server.Events, 
 		result = append(result, event)
 	}
 
+	// rows.Next() returns false both at the end of the result set and when
+	// the query dies mid-flight (killed session, cancelled statement, broken
+	// connection). Without surfacing rows.Err(), an interrupted query is
+	// indistinguishable from an empty result, which callers interpret as "key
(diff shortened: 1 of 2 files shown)


nic-6443:
Closing — not pursuing this from my side.

brandond:
> I hit this in production-like testing on the SQL Server backend

We don't have a Microsoft SQL server backend in this repo... is this issue perhaps unique to whatever fork you're using?

PR #739, opened 2026-08-12
Update to etcd 3.7 / Kubernetes 1.37 / golang 1.26

brandond:
This will be required for K3s 1.37, as k8s 1.37 calls `RangeStream` which is only in etcd 3.7 - even if the server just returns Unimplemented, we will still need to update the protobuf definitions for it to build.

See also: https://etcd.io/docs/v3.7/learning/api/#rangestream


brandond:
Leaving this in draft until after Kubernetes 1.37 is out of RC

matttrach (approved):

vitorsavian (approved):

dereknola (approved):

Issue #731, opened 2026-08-01
Watch progress messages may send current revision before events for revision are observed


brandond:
The current method for generating watch progress messages (WatchResponse messages with no events, that indicate to the client the current revision counter) are unreliable.

Progress messages MUST NOT send a revision before all events for that revision have been sent to the client. If the client has not seen all events, a progress WatchResponse MUST NOT be sent. Sending a progress notification before the event for that revision will cause the client watcher (apiserver) to drop the event, and permanently desync its caches.

Unfortunately, due to the several layers of buffered channels between the actual backend poll loop, and the GRPC watch server, it is impossible to know if events are queued for a client. The poll loop knows what revisions it has sent into the channel, and the watcher knows what the backend's current revision is. We currently wait for the two to converge (via `WaitForSyncTo()`) before sending progress notifications, but this is not sufficient. Under heavy load, the watcher may not receive messages from the channel until tens of milliseconds after they are sent. When this occurs the progress notification and events occur out of order, and the client becomes desynced.

Example from trace logs from a server with the changes from https://github.com/k3s-io/kine/pull/728:

```
TRAC[2026-08-01T02:39:26.885784491Z] POLL AFTER 65608, limit=500, events=2
TRAC[2026-08-01T02:39:26.885847496Z] BROADCAST /registry/jobs/fleet-local/test-gitrepo-5eeee, revision=65609, delete=false
TRAC[2026-08-01T02:39:26.885864204Z] BROADCAST /registry/fleet.cattle.io/gitrepos/fleet-local/test-gitrepo, revision=65610, delete=false

TRAC[2026-08-01T02:39:26.886480838Z] WATCH REQUEST PROGRESS server=183
TRAC[2026-08-01T02:39:26.887506607Z] WATCH PROGRESS TICK server=20
TRAC[2026-08-01T02:39:26.887543614Z] WATCH SEND server=18, id=18, key=/registry/jobs/, revision=65609, events=1, size=11923, reads=1, keys=[/registry/jobs/fleet-local/test-gitrepo-5eeee]
TRAC[2026-08-01T02:39:26.887702869Z] WATCH SEND PROGRESS server=183, id=-1, revision=65610
TRAC[2026-08-01T02:39:26.88779876Z] STREAM STATS WATCH SEND DONE client=172.17.0.4, id=-1, revision=65610, events=0, size=17, created=false, canceled=false, reason="", time=49µs
[...]

PR #746, opened 2026-08-20
Watch event pipeline refactor

brandond:
Major refactor of the watch event pipeline, along the lines of the plain laid out in:
* https://github.com/k3s-io/kine/issues/731

Todo:
* [x] Backends must implement a single watcher that observes events for all keys.
* [x] Backends must use the existing broadcaster package to fan-out events to multiple client watchers, without filtering events by key prefix.
* [x] The watch server must see all events from the backend, and perform key range filtering based on the client's requested range.
* [x] The watch server must track the max revision that each watch has seen, where "seen" means the event was either dropped due to not matching the range, or the event has been enqueued for send to the client if it did match the range.
* [x] The watch server must only send progress notification messages to a watch if its max seen revision is equal to the backend's current revision. A watcher that has seen the current revision is "synced" in etcd parlance.
* [x] The watch server must not block the event channels; it should always read all events from the channel
* [ ] ~The watch server should drop GRPC clients that fall behind on receiving messages. Etcd uses a WatchResponse message queue for this, and drops clients if the queue reaches a maximum size, or if the send fails.~
  *Leaving this unhandled for now, as it is not something we have done in the past, nor has it caused any problems.*

In addition to actually guaranteeing that watchers are synced before sending progress notifications (instead of simply checking that the poll loop has sent events for a revision into the channel), this also does away with the 100 millisecond delay that was artificially added to progress notification responses. This plus the 1 second poll interval was a cause of significant slowness (0.1 to 1.1 seconds * number of namespaced resource types present in the cluster) when the apiserver requested progress notifications to ensure that all resources are deleted from a namespace. (ref https://github.com/k3s-io/k3s/issues/14399)


vitorsavian (approved):

manuelbuil (approved):
Looks very good Brad. It addresses all the points in the issue and the logging looks good for debugging revisions,etc. My only concern is that, If I understand correctly, messages will now stay a bit longer inside Kine before being filtered out. I was wondering if that means any relevant extra CPU/Memory requirement. I guess not

brandond:
@manuelbuil they make it through an additional channel before being filtered out now yes, but since all receivers get the same event list there's no additional memory utilization. Just a bit more work for the final goroutine that is now seeing more than it used to.

Issue #676, opened 2026-05-26
Potential event loss due to MySQL auto-increment commit ordering gap


chenk008:
## Summary

When using MySQL as the backend, there is a potential data consistency issue where a Watch client can miss events due to the gap between auto-increment ID allocation and transaction commit visibility.

## Root Cause

MySQL InnoDB allocates auto-increment IDs at INSERT time, but rows only become visible after COMMIT. Transactions may commit out of order:

```
TX1: INSERT /a → gets id=100 → slow transaction...
TX2: INSERT /b → gets id=101 → COMMIT ✓ (id=101 now visible)
```

At this point, `MAX(id)` returns 101, but id=100 is not yet visible.

## Scenario

1. **Client calls `List(revision=0)`**:
   - SQL: `SELECT MAX(rkv.id) FROM kine` → returns `rev=101`
   - Data result does not include the change at id=100 (TX1 uncommitted)
   - Client receives `rev=101`

2. **Client starts `Watch(revision=101)`**:
   - In `logstructured.Watch()`, revision is decremented to 100
   - `After(prefix, 100, 0)` executes `SELECT ... WHERE kv.id > 100`
   - If TX1 is still uncommitted, After only returns id=101, and `rev=101` (from MAX(id))
   - `lastRevision` is set to 101

3. **TX1 finally commits** (id=100 becomes visible):
   - The internal `poll()` goroutine detects id=100 via gap-fill mechanism
   - Event id=100 is delivered to `readChan`
   - However, `filter(events, lastRevision)` discards events where `ModRevision <= lastRevision`
   - Since `100 <= 101`, **the event at id=100 is silently dropped**

4. **Result**: The change at id=100 is never observed by the client — not in List results, not via Watch.

## Relevant Code

- `List()` returns `rev` from `MAX(id)`
- Watch decrements revision by 1
- `After()` query with decremented revision
- filter()` drops events ≤ lastRevision
- poll gap-fill mechanism (only protects Watch event ordering, not the After+filter interaction)

## Impact

- In practice, the window is very small because single-row INSERT transactions in MySQL commit quickly
- If `LockWrites` is enabled, write serialization prevents this issue
- Kubernetes clients have relist mechanisms that will eventually resync
- However, under high-concurrency write workloads on MySQL without `LockWrites`, this is a theoretically reachable consistency gap

## Possible Mitigations

[...]

PR #759, opened 2026-09-04
fix(sql): Avoid stalled poll loop in high-write environments

thomastaylor312:
**UPDATE**: While the high write speed I initially found below does trigger the issue, it probably isn't what people are actually running into. The real issue with my production was gaps introduced due to conflicted writes and `SERIAL` types in Postgres causing gaps. I've updated the reproduction in the repo linked below

This is an attempt to solve a bug I found running k3s in a production environment with high writes.

I wrote up a longer form description of the issue, along with instructions on how to reproduce the issue here: https://github.com/thomastaylor312/kine-repro. This is taken directly from the README there so it is available in the issue too:

## What it looks like

This was hard to diagnose because almost everything still works. But this seemed to be the pattern:

- `apiserver_watch_cache_events_received_total` goes flat for **every** resource, on **every** replica, at roughly the same time.
- Writes keep committing at full rate.
- `kubectl get` still returns current data, because a quorum read goes straight to the datastore and never touches the watch cache.
- Controllers and operators quietly stop reacting to anything. Their informers are alive, connected, and receiving nothing.
- It never recovers on its own. Restarting the process fixes it instantly and completely.

## What I think is happening

TL;DR: kine's compactor deletes rows that its own poll loop has not read yet, and the repair path for the resulting hole is much slower than normal operation, so it can never catch back up.

The longer version, in order:

1. `poll()` in `pkg/logstructured/sqllog/sql.go` is a single goroutine that queries all keys and is the only source of watch events for every prefix on the process. There is one cursor, and everything downstream is fed from it.
2. Under sustained write load that cursor falls behind the write revision. Under normal operation it catches up in batches of `pollBatchSize` (500–1000).
3. The compactor is driven entirely by the write revision. `safeCompactRev` compares against `currentRev`, and `polledRev` is referenced in exactly three places in the whole codebase: where it's declared, where `poll()` writes it, and where `WaitForSyncTo` reads it. Nothing in the compaction path ever looks at it. So compaction happily deletes rows the poll loop has not consumed.
4. Now `poll()` queries for `pollRevision+1` and gets back a row with a higher `ModRevision`. It sees a hole. That hole is permanent, because the row it's waiting for was deleted.
5. The gap branch calls `Fill()`, which inserts a `gap-<rev>` marker row and advances the cursor by _exactly one revision_ before starting on the next loop
6. So the poll loop is now repairing at one revision per query-plus-insert round trip instead of draining hundreds per query. I measured roughly 27 revisions/second in my production workload.

Once the repair rate drops below the write rate (which was around 60-70/s in my production workload), the gap diverges and the poll loop never emits another real event. For all intents and purposes, every watch on that process is dead.

There is a `canSkipRevision` that will give up on a hole after one second and skip forward, logging a `GAP` error. But it only fires when `Fill` keeps *failing*. On the success path, filling advances the cursor by one, which changes the revision being waited on, which resets the tracking. So the one-second timer never expires. You end up with zero `GAP` lines. 

> [!NOTE]
> - I completely understand that this is fairly stable code and is considered mostly correct. I found the other issues that touched this area already, so maybe what is going on is already the right answer and this fix is not the right way to solve it. Either way, I thought this would be useful
> - I have no problems if this is completely thrown away, I just thought the problem would be easier to show via code + my repro repository. I can also close this PR and open as just an issue
> - I also know #746 is in progress, but as far as I could tell, it doesn't address this issue
> - AI helped me write the repro, but I have read through every line and basically re-wrote the readme.

### DIFF
--- a/pkg/logstructured/sqllog/sql.go
+++ b/pkg/logstructured/sqllog/sql.go
@@ -18,16 +18,45 @@ import (
 
 const minCompactBatchSize = 100
 
+// maxPollStall is how long the poll loop may go without advancing before compaction stops holding
+// its target back to the poll cursor. Under load the cursor should advance continuously, so this
+// limit wouldn't be hit. When the datastore is idle it sits at the current revision and the limit
+// isn't needed either. If it stops moving for this long the poll loop is assumed to be wedged for
+// some unrelated reason, and compaction is allowed to proceed so that the database does not grow
+// unbounded. The poll loop will then hit a compacted revision and restart the watch stream,
+// matching what happens in the etcd-based code in k8s.io/apiserver.
+const maxPollStall = time.Minute
+
+// maxCompactHold bounds how far compaction may be held back from the revision it would otherwise
+// compact to, while waiting on the poll loop. We need both this and maxPollStall because, under
+// sustained overload the poll cursor keeps advancing, just more slowly than the write head, so the
+// hold would grow without limit and take the size of the database with it. Past this distance the
+// reader is not likely going to catch up, so compaction proceeds anyway and the poll loop takes the
+// same compacted-revision restart it would have taken had we never held it back.
+const maxCompactHold = 10000
+
 type SQLLog struct {
 	sync.RWMutex
 
-	d                     server.Dialect
-	broadcaster           broadcaster.Broadcaster
-	ctx                   context.Context
-	notify                chan int64
-	currentRev            atomic.Int64
-	polledRev             atomic.Int64
-	polled                *sync.Cond
+	d           server.Dialect
+	broadcaster broadcaster.Broadcaster
+	ctx         context.Context
+	notify      chan int64
+	currentRev  atomic.Int64
+	polledRev   atomic.Int64
+	polled      *sync.Cond
+
+	// polling is true while the poll loop is running; polledAt records the last time polledRev
+	// actually advanced. Compaction consults both so that it does not delete rows the poll loop has
+	// not read yet. See compact().
+	polling  atomic.Bool
+	polledAt atomic.Int64
+
+	// compactorOnce guards the compactor goroutine. The poll loop can now exit
+	// and be restarted by the broadcaster, which calls startWatch again; without
+	// this we would leak an additional compactor on every restart.
+	compactorOnce sync.Once
+
 	compactInterval       time.Duration
 	compactIntervalJitter int
 	compactTimeout        time.Duration
@@ -256,6 +285,21 @@ func (s *SQLLog) compact(compactRev int64, targetCompactRev int64) (int64, int64
 	// Ensure that we never compact the most recent 1000 revisions
 	targetCompactRev = safeCompactRev(targetCompactRev, currentRev, s.compactMinRetain)
 
+	// Never compact past the poll loop's position while it is making progress. The poll loop is not
+	// a client watcher that can be told to resync. It is this process' _only_ reader of the log,
+	// and every watch is fed from it. Deleting rows it has not read yet leaves a permanent hole
+	// that cannot be filled, causing a hang. See the actual poll() loop for why this is a problem
+	// and how it is handled.
+	if s.polling.Load() && time.Since(time.Unix(0, s.polledAt.Load())) < maxPollStall {
+		if polledRev := s.polledRev.Load(); polledRev > 0 && targetCompactRev > polledRev {
+			// Hold back to the poll cursor, but never by more than maxCompactHold.
+			if heldRev := max(polledRev, targetCompactRev-maxCompactHold); heldRev < targetCompactRev {
+				logrus.Debugf("COMPACT holding target revision %d back to %d for poll cursor %d", targetCompactRev, heldRev, polledRev)
+				targetCompactRev = heldRev
+			}
+		}
+	}
+
 	// Don't bother compacting to a revision that has already been compacted
 	if targetCompactRev <= compactRev {
 		logrus.Tracef("COMPACT revision %d has already been compacted", targetCompactRev)
@@ -462,12 +506,15 @@ func (s *SQLLog) startWatch() (chan server.Events, error) {
 	maxJitter := float64(s.compactIntervalJitter) / 100.0 * float64(s.compactInterval)
 	jitter := time.Duration(rand.Float64()*2*maxJitter - maxJitter)
 
-	// start compaction and polling at the same time to watch starts
-	// at the oldest revision, but compaction doesn't create gaps
+	// The compactor is started only once for the lifetime of the SQLLog. The poll loop may exit and
+	// be restarted (see poll()), which brings us back through startWatch, and a second compactor
+	// would double-compact.
 	if s.compactInterval <= 0 {
 		logrus.Debugf("COMPACT disabled; automatic compaction will not occur")
 	} else {
-		go s.compactor(s.compactInterval + jitter)
+		s.compactorOnce.Do(func() {
+			go s.compactor(s.compactInterval + jitter)
+		})
 	}
 
 	go s.poll(c, pollStart)
@@ -480,12 +527,19 @@ func (s *SQLLog) poll(result chan server.Events, pollStart int64) {
 		skipTime     time.Time
 		waitForMore  = true
 		pollRevision = pollStart
+		lastPolled   = pollStart
 		trace        = logrus.IsLevelEnabled(logrus.TraceLevel)
 	)
 
 	wait := time.NewTicker(time.Second)
 	defer wait.Stop()
 	defer close(result)
+
+	// This is where we mark that the poll loop is running and record the last time it advanced. The
+	// compactor consults both of these to avoid deleting rows that the poll loop has not yet read.
+	s.polling.Store(true)
+	s.polledAt.Store(time.Now().UnixNano())
+	defer s.polling.Store(false)
 
 	for {
 		if waitForMore {
@@ -503,6 +557,10 @@ func (s *SQLLog) poll(result chan server.Events, pollStart int64) {
 
 		//  update polled revision to reflect what rows have already been seen
 		s.Lock()
+		if pollRevision != lastPolled {
+			lastPolled = pollRevision
+			s.polledAt.Store(time.Now().UnixNano())
+		}
 		s.polledRev.Store(pollRevision)
 		s.polled.Broadcast()
 		s.Unlock()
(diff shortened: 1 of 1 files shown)


thomastaylor312:
I did test this fix with a manual build of this patch against 0.16, built into k3s and it did seem to solve the issue when I ran my repro script against it

pkg/logstructured/sqllog/sql.go
```diff
@@ -18,16 +18,45 @@ import (
...
+// matching what happens in the etcd-based code in k8s.io/apiserver.
+const maxPollStall = time.Minute
+
+// maxCompactHold bounds how far compaction may be held back from the revision it would otherwise
+// compact to, while waiting on the poll loop. We need both this and maxPollStall because, under
+// sustained overload the poll cursor keeps advancing, just more slowly than the write head, so the
+// hold would grow without limit and take the size of the database with it. Past this distance the
+// reader is not likely going to catch up, so compaction proceeds anyway and the poll loop takes the
+// same compacted-revision restart it would have taken had we never held it back.
+const maxCompactHold = 10000
```
thomastaylor312, line 36: I kinda picked these out of a hat based on the kind of load I was seeing in my cluster. These might either need to be different and/or configurable

pkg/logstructured/sqllog/sql.go
```diff
@@ -256,6 +285,21 @@ func (s *SQLLog) compact(compactRev int64, targetCompactRev int64) (int64, int64
 	// Ensure that we never compact the most recent 1000 revisions
 	targetCompactRev = safeCompactRev(targetCompactRev, currentRev, s.compactMinRetain)
 
+	// Never compact past the poll loop's position while it is making progress. The poll loop is not
```
brandond, line 288: this is what compactMinRetain is for... if the poller is further behind than the target (when adjusted by compactMinRetain) then something has gone terribly wrong.

Note that kine instances do not communicate with each other, so just because **this** poller has caught up to a particular revision, doesn't mean others have.  We assume that no instance ever gets more than compactMinRetain revisions behind, but do not actually check for that.

brandond:
I am really confused as to how you would even run into the problem you are describing. compact-min-retain defaults to 1000. How are you managing to tank performance so badly that the poller lags that far behind, AND the apiserver tries to compact into that window? The apiserver compacts to what the current revision was 5 minutes ago, and compactMinRetain ensures that at least 1000 rows are always retained for the poller to catch up. It seems like you would need to be compacting at abnormally short intervals, and have turned down the compact-min-retain setting, in order to hit this?

brandond:
I will also note that now is not a GREAT time for this contribution, as I am in the process of rewriting the entire watch pipeline in https://github.com/k3s-io/kine/pull/746 in order to address a couple long-standing issues with watch progress notifications. If whatever's going on in your environment isn't totally bonkers, and is still an issue once that's done, see if you can rebase your changes on top of that?

brandond:
> What I think is happening

I don't agree with your hypothesis here.

I think that what is probably ACTUALLY happening is what I am trying to fix in the other PR, and is described in https://github.com/k3s-io/kine/issues/731 - the apiserver caches are seeing a watch progress notification with an out-of-sequence currentRevision, and their caches become permanently desynced.

You can test this theory by running your apiserver (or k3s) with `KUBE_WATCHCACHE_CONSISTENCY_CHECKER=true` - it will check the cache against the datastore periodically, and panic if they are out of sync.

I would want to see that NOT failing, and have some logs from kine running at debug level (you'll need to run it standalone to get full trace-level query logging) showing the poller being further behind than the compact target, in order to believe what you're asserting is happening.

thomastaylor312:
> I will also note that now is not a GREAT time for this contribution, as I am in the process of rewriting the entire watch pipeline in https://github.com/k3s-io/kine/pull/746 in order to address a couple long-standing issues with watch progress notifications. If whatever's going on in your environment isn't totally bonkers, and is still an issue once that's done, see if you can rebase your changes on top of that?

For sure! This is why I called #746 out in my initial description as I knew that would probably change things. I just wanted to capture something that seemed to help, but wasn't entirely convinced because all of this code is entirely new to me and figured it was likely wrong. The more important thing was the reproduction that seemed to follow the pattern I was seeing with some of our production workloads. As to your point about compaction intervals, we haven't tweaked anything lower (I did it lower in my reproduction to help trigger the issue sooner, though I can try for longer), though we have experimented with making it higher, so I don't think anything is weird there. I will say the number of writes/s in this cluster is pretty large. We had k3s running 2 instances with the following amount of resources:

```yaml
resources:
    limits:
      memory: 12Gi
    requests:
      cpu: "6"
      memory: 12Gi
```

These instances have the following write rates, so I'm wondering if they fall into your definition of "bonkers" 😆 

Total:
Screenshot 2026-09-08 at 09 30 56

Writes/s
Screenshot 2026-09-08 at 09 33 36

> You can test this theory by running your apiserver (or k3s) with KUBE_WATCHCACHE_CONSISTENCY_CHECKER=true - it will check the cache against the datastore periodically, and panic if they are out of sync.

TIL about this env var. I'll give it a whirl at least on my repro repo for now and see if we can try it some places in our live prod and then report back!

thomastaylor312:
I'm currently trying to make my repro match what I've got in prod as closely as possible now since I was just trying to get it to the point that the actual cache tips over into wedging. As I've been running this a few more times and combing through the output, I think that my original observation about the repair rate being a single row at a time (the gap filling logic) means if that gets to be slower than the write speed (about 55-66 writes/s in the example from my prod workload), then the issue I observed starts happening. What I am trying to figure out is what is causing that gap. 

I think one of the reasons we might be getting gaps is because the resource version is a serial column. Based on what I was seeing and my read of how a `SERIAL` value works, a failed insert/transaction doesn't rollback the value itself. It leads to gaps instead. So with 2 k3s instances writing fairly heavily at the same time, I could see gaps happening. When I ran my test I saw plenty of 

```
time="2026-09-08T18:37:09Z" level=error msg="Insert error for key /registry/configmaps/loadtest/obj-0067: ERROR: duplicate key value violates unique constraint \"kine_name_prev_revision_uindex\" (SQLSTATE 23505)" createRevision=685 key=/registry/configmaps/loadtest/obj-0067 previousRevision=10689
```

Which would cause an ID to be skipped, leading to a gap. Then, once we can't recover, we'd eventually make the problem get worse with the compaction firing off. I went back to our production logs and found that right before we started having the issue, we saw this line show up about 2600 times within a minute (along with a continuous write speed of 55 w/s), which I think tipped the gap over into being large enough. In our specific case, it looks like there was a conflict when trying to create a whole bunch of events on a single kubernetes object.

So I think that is why we're getting into the situation we are getting into. What I am going to do next is update my repro to generate conflicts and see if I can get into the same state. I'll make sure I have `KUBE_WATCHCACHE_CONSISTENCY_CHECKER` on there as well just to make sure and I'll report back again

brandond:
Yes. Postgres is kind of shitty in how it implements auto-increment keys via SERIAL... even if the insert fails due to a unique constraint, the sequence number is consumed and you end up with a gap that we fill to ensure strict revision stepping. This isn't an issue with any other engine.

Any time you have a bunch of clients racing to modify a resource, you're likely to end up with a conflict on the insert due to multiple writes trying to insert a new row with the same name and prev_revision, which must be unique. On postgres this is guaranteed to create a gap.

brandond:
See also https://github.com/k3s-io/kine/issues/676#issuecomment-5038261320

thomastaylor312:
I've got the repro working with something that generates a lot of conflicts and triggers this behavior. Going to finish cleaning up tomorrow and push it, but it boils down to:

1. Have load or a storm that triggers a bunch of conflicts, which leads to a gap in IDs
2. The `Fill()` function only inserts 1 row at a time, including latency to database calls. In my local tests with 3ms of injected latency, this held the fill rate to ~25 revisions/s
3. If your write rate is above 25/s, then you're never going to catch up

That is my working theory, but I want to get my repro up so you can see what I did as well

brandond:
I'd recommend you take a look at the other thread. There is an outstanding question of how long we should reasonably wait around for a transaction to become visible, and what to do if we are unwilling to wait any longer.

I'd also avoid making too many changes to this PR before I get the watch rewrite merged; I'm not going to accept anything structural until that's done.

thomastaylor312:
Yeah I wasn't looking to change here since you have the other work in flight. I just want to confirm the behavior for sure so we can use that as a basis for discussion. 

I did read that other comment and I'm not seeing those `GAP` log lines anywhere in my repro or the production logs, so it isn't reaching that part of the code at least. But I am seeing the slowness that comes from a combination of the delay before filling plus the external latency. Problem is, I haven't been in this problem space for long enough to have an opinion that is of much value 😅 (I mostly know what we do in our case). I do know we've started seeing this happen several times, which is why I brought it up. But I fully admit we could be a special snowflake with either our postgres setup or how we're (ab)using k3s and kine. 

Obviously Postgres makes this more complicated due to the nature of SERIAL types adding bigger gaps. The fix I attempted here in the PR in the first place was to just have the loop exit early so that it would return `ErrCompacted` to the clients and force the poll loop to restart. I don't know if that is a terrible idea for everyone else outside postgres though. That seems to be what etcd does, which seems like a decent path to follow, but I don't know if that is the best thing to do for Kine itself

To put it another way, I only have my experience here and don't want to come in demanding a solution just because it works for me. I'm also fine to revisit this after your other refactor as well. Just trying to be helpful!

thomastaylor312:
Ok, I've updated my reproduction at https://github.com/thomastaylor312/kine-repro so that I have a way to reliably introduce the issue. Now that I've stared at the code for long enough, I have a few ideas. First and foremost is waiting for #746 to land and then see if I can still reproduce the issue. At that point, there are a couple of ideas I've had (not all fleshed out) of what the next steps could be:

- Do nothing and document this as mostly a Postgres issue due to the skipped numbers from `BIGSERIAL`. Technically the bug still exists, but it requires extremely high write speeds or something else to cause an initial gap in other providers
- Take some sort of version of what I did in this PR of restarting the poll loop and causing current watches to return an error
- Add some additional bookkeeping to track how far behind the poll's read revision is behind (maybe if it is far enough behind the compaction interval) and then reset the poll or jump forward

I have no idea if any of these would be good or just cause issues for other drivers, but wanted to get them out there regardless. 

But all of this should wait until after #746 lands for now

brandond:
v0.17.1 is out with some major changes. Give it a look if you would.

PR #760, opened 2026-09-07
Migrate from urfave/cli v2 to v3

mrueg:
## Problem

`github.com/urfave/cli/v2` is in maintenance mode; v3 is the actively developed line.

## Solution

Move `pkg/app` to `github.com/urfave/cli/v3` (v3.11.0). The API changes applied:

- `cli.App` / `cli.NewApp()` → a `cli.Command` struct literal
- flag `EnvVars: []string{...}` → `Sources: cli.EnvVars(...)`
- action signature `func(*cli.Context) error` → `func(context.Context, *cli.Command) error`
- `Run(args)` → `Run(ctx, args)`

`app.New()` now returns `*cli.Command` instead of `*cli.App` — the only signature change visible to importers of `pkg/app`. `app.Config()` is unchanged.

As a side effect, v3 drops three indirect dependencies that only the v2 docs generator pulled in: `cpuguy83/go-md2man/v2`, `russross/blackfriday/v2` and `xrash/smetrics`.

## Verification

- `go build ./...`, plus `CGO_ENABLED=0 go build ./...` and `go build -tags nats ./...`
- `go test -tags=test ./...` — all packages pass
- `golangci-lint run` — no new findings
- `--help` output, flag defaults, env var names and `--version` are byte-for-byte equivalent to v2; `kine foo` still fails with `kine does not accept positional arguments, only flags`, and an unknown flag still prints usage and exits 1
- Started the binary against sqlite with `KINE_ENDPOINT` / `KINE_LISTEN_ADDRESS` / `KINE_METRICS_BIND_ADDRESS` set to confirm env-var sources still bind, and that shutdown on SIGTERM is unaffected


brandond (changes requested):
Thanks, but we're still on v2 for k3s and rke2; all projects in the ecosystem should move to v3 in the same release cycle.

https://github.com/k3s-io/k3s/blob/v1.36.4%2Bk3s1/go.mod#L142

I'm not seeing anything that would convince me that there is a pressing need to get off v2.

dereknola:
As the one who did the bulk of the `v1` to `v2` migration, we cannot easily move to v3 until containerd, and crictl moves to v3. 

Additionally urfave as continued to make substantial changes to how the `cli` library works, which often leads to feature gaps or "it just works differently now" between version. 

Containerd *just* moved to `v3` 2 days ago, https://github.com/containerd/containerd/pull/14095, but its not in any release avalible yet. cri-tools is still on urfave v2. 

As Brad said, there aren't any CVEs in v2 and there are not any new features we are currently needed that v3 provides over v2, so our drive to move to v3 is low.

mrueg:
Thanks for the feedback, appreciate it! I understand the context now better, that you prefer aligning that across all rancher tools. 

And thanks again for @brandond for reviewing the other PRs so quickly!

Closing this one then :)

PR #764, opened 2026-09-07
pgsql: push the list row limit into the key selection

mrueg:
## Problem

The portable list query (`generic.ListFmt`) joins `kine` against a subquery that aggregates the current revision of every key in the range:

```sql
SELECT current_rev, compact_rev, <cols>
FROM (SELECT MAX(id) FROM kine) AS current, (…) AS compact, kine
INNER JOIN (SELECT MAX(id) AS id FROM kine
            WHERE name >= $1 AND name < $2 GROUP BY name) AS mkv USING (id)
WHERE (deleted = 0 OR $3)
ORDER BY name ASC LIMIT 501
```

The `LIMIT` sits outside everything, so **it bounds the rows returned, not the rows scanned**. The `GROUP BY` must aggregate every historical revision of every key in the range — not just the live keys — and the `deleted` filter is applied after that, before the outer sort can truncate. Since the apiserver's chunked reflector sync pages at limit=500, that whole aggregation is repeated for every page.

The `kine_list_query_index (name, id DESC, deleted)` that the pgsql driver creates specifically for this query can never be exploited by that shape: `MAX(id) … GROUP BY name` plans as a GroupAggregate, which reads every index entry and cannot use the `id DESC` ordering to stop early.

## Solution

On Postgres the same result is expressible as `DISTINCT ON (name) … ORDER BY name ASC, id DESC`, which keeps the whole row available. That lets the `deleted` filter run **ahead** of the limit — preserving the generic query's semantics — and lets the scan stop as soon as the limit is satisfied.

Applying a limit inside a subquery requires placing it somewhere other than the end of the statement, so `query.Named` gains `WithLimit`: queries carrying `query.LimitToken` have it substituted, queries without it get the limit appended exactly as before. The token is written as an SQL comment (`/*limit*/`) so a query carrying it remains valid SQL if the limit is never applied. The three list call sites in `generic` route through it; the `After()` queries keep appending, since they have no subquery to push into.

Only the pgsql driver overrides the queries. The generic layer, SQLite and MySQL are unaffected — `DISTINCT ON` is Postgres-specific, and there is no portable rewrite that keeps the `deleted` filter ahead of the limit without window functions.

## Measurements

PostgreSQL 17.11, table built from this driver's exact schema: 300,000 rows, 20,000 distinct keys, 392 MB, ~6% tombstones. Listing a 20,000-key range:

| | run 1 | 2 | 3 | 4 | 5 |
|---|---|---|---|---|---|
| limit 501, current | 128.0 | 128.1 | 131.1 | 127.9 | 127.0 ms |
| limit 501, pushed down | 7.0 | 6.1 | 6.2 | 6.2 | 6.2 ms |
| no limit (18,823 rows), current | 163.3 | 159.6 | 155.5 | | |
| no limit, pushed down | 139.5 | 117.3 | 119.0 | | |

The unlimited case is included deliberately, since it has no limit to push down — it does not regress.

The plan changes from a parallel index scan over all 300,000 index entries plus a top-N sort:

```
Limit  Buffers: shared hit=128479 read=38418
  ->  GroupAggregate  ->  Index Only Scan using kine_name_id_index (actual rows=300000)
Execution Time: 213.564 ms
```

to an index-only scan that stops early:

```
Limit (actual rows=501)  Buffers: shared hit=1169
  ->  Unique  ->  Index Only Scan using kine_list_query_index (actual rows=7966)
                    Heap Fetches: 0
Execution Time: 2.826 ms
```

## Correctness

The two query shapes were diffed **row for row** against the same database across limits 0/1/2/501/5000/20000/30000, both values of the `deleted` flag, three revision bounds, an empty range and a single-key range — 58 comparisons, all identical. That includes the boundary the rewrite has to get right: where the `deleted` filter reduces the result below the requested limit (18,823 rows for a limit of 20,000).

End to end, a paginated LIST through the etcd API (2,000 keys, 3 deleted, paging at limit 500 the way the reflector does) returns byte-identical results on this branch and on master:

```
pages=4 counts=[1997 1497 997 497] paged_keys=1997 unpaged_keys=1997 first=key-00000 last=key-01998
```

with paged and unpaged results matching each other key for key, in order, on both builds.

`go test -tags=test ./...` passes; `golangci-lint run` reports only the pre-existing `pkg/util` package-naming finding. A unit test for `WithLimit` covers substitution, appending, and both no-limit paths.

## Caveats

- **Visibility map.** The `Heap Fetches: 0` above is on a freshly vacuumed table. On a continuously-inserted table whose pages are not all-visible, the index-only scan degrades to an index scan with heap fetches. The push-down still wins — it visits far fewer rows — but the headline ratio will be smaller in a busy cluster.
- **CockroachDB** is a supported target for this driver and does support `DISTINCT ON`, but I have no CockroachDB instance to test against, so that path is unverified.
- **Counts are untouched.** `CountCurrentSQL`/`CountRevisionSQL` still aggregate the whole range, and `list.go` still issues one per page. Rewriting them the same way measured only 84 ms → 66 ms, because a count genuinely has to visit every key; the real fix there is to stop recomputing it per page, which changes `RangeResponse.Count` semantics and belongs in its own change.


brandond:
Can you test with cockroachdb to see how this works on it, and confirm it doesn't regress anything? If you (or your pet bot) can spin up a postgres DB you should be able to spin up cockroachdb as well. There are examples in the test harness.

pkg/drivers/pgsql/pgsql.go
```diff
@@ -54,6 +54,38 @@ var (
 		`CREATE UNIQUE INDEX IF NOT EXISTS kine_name_prev_revision_uindex ON kine (name, prev_revision)`,
 		`CREATE INDEX IF NOT EXISTS kine_list_query_index on kine(name, id DESC, deleted)`,
 	}
+	// listFmt selects the current revision of every key in a range, in name
```
brandond: I know LLMs love to add verbose expository comments, but this code base does not do that. None of the other queries have this sort of commentary. Remove it from here, and save it for the commit message or PR description.

pkg/drivers/pgsql/pgsql.go
```diff
@@ -54,6 +54,38 @@ var (
...
+			SELECT dkv.id AS list_id
+			FROM (%s) AS dkv
+			WHERE (dkv.deleted = 0 OR ?)
+			ORDER BY dkv.name ASC
+			` + query.LimitToken + `
+		) AS mkv
+		JOIN kine ON kine.id = mkv.list_id
+		ORDER BY kine.name ASC`
+
+	// distinctNameSQL selects the highest id -- the current revision -- of every
```
brandond: Emdashes aren't any less of an LLM tell when they're in comments as `--`...  Drop this comment as well.

pkg/query/query_test.go
brandond, line 1: Drop this file, see: https://github.com/k3s-io/k3s/blob/main/CONTRIBUTING.md#ai-guidance. This code base prefers to test code by using it, not with synthetic tests that exercise inputs and outputs in isolation.

pkg/drivers/generic/generic.go
```diff
@@ -382,9 +382,7 @@ func (d *Generic) ListCurrent(ctx context.Context, key, end string, limit int64,
 	} else {
 		sql = d.ListCurrentValSQL
 	}
-	if limit > 0 {
-		sql = sql.Appendf("LIMIT %d", limit)
-	}
+	sql = sql.WithLimit(limit)
```
brandond, line 385: With these changes, are there even any other uses of Appendf? If not, it can be removed from the SQL package.

brandond (changes requested):

pkg/query/query.go
```diff
@@ -39,6 +48,26 @@ func (n *Named) Appendf(format string, a ...any) *Named {
 	return &Named{Name: n.Name, Query: n.Query + " " + fmt.Sprintf(format, a...)}
 }
 
+// WithLimit returns a copy of the named query with a row limit applied.
+//
+// Queries containing LimitToken have it replaced by the limit clause, which
+// lets a driver position the limit inside a subquery instead of at the end of
+// the statement. Queries without the token get the limit appended, as before.
+// A limit of zero or less applies no limit, and only strips the token.
```
brandond: "as before"? Before what?

Too much talking. Just leave this out.

```suggestion
```

brandond (changes requested):

pkg/drivers/pgsql/pgsql.go
```diff
@@ -54,6 +54,38 @@ var (
...
+	// Format args: current revision, compact revision, columns, key selection.
+	listFmt = `
+		SELECT (%s) AS current_rev, (%s) AS compact_rev, %s
+		FROM (
+			SELECT dkv.id AS list_id
+			FROM (%s) AS dkv
+			WHERE (dkv.deleted = 0 OR ?)
+			ORDER BY dkv.name ASC
+			` + query.LimitToken + `
+		) AS mkv
```
brandond, line 65: Ok so the idea here is force postgres to use the index on `name, id DESC, deleted` by adding a subquery that uses only those rows, and then limiting prior to joining. I don't love that this ends up nesting subqueries three levels deep, but if thats what it takes to hint to the postgres planner that it should use the index I guess it makes sense.

brandond:
Before:
```
                                                                      QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=586.37..586.64 rows=108 width=546)
   Sort Key: kine.name COLLATE "C"
   ->  Nested Loop  (cost=9.20..582.72 rows=108 width=546)
         ->  Nested Loop  (cost=8.92..118.32 rows=108 width=24)
               ->  Nested Loop  (cost=8.64..8.67 rows=1 width=16)
                     ->  Result  (cost=0.34..0.35 rows=1 width=8)
                           InitPlan 1
                             ->  Limit  (cost=0.28..0.34 rows=1 width=8)
                                   ->  Index Only Scan Backward using kine_pkey on kine kine_2  (cost=0.28..206.35 rows=3580 width=8)
                     ->  Result  (cost=8.30..8.31 rows=1 width=8)
                           InitPlan 2
                             ->  Limit  (cost=0.28..8.30 rows=1 width=8)
                                   ->  Index Only Scan Backward using kine_name_prev_revision_uindex on kine kine_3  (cost=0.28..8.30 rows=1 width=8)
                                         Index Cond: ((name = 'compact_rev_key'::text) AND (prev_revision IS NOT NULL))
               ->  GroupAggregate  (cost=0.28..107.49 rows=108 width=60)
                     Group Key: kine_1.name
                     ->  Index Only Scan using kine_list_query_index on kine kine_1  (cost=0.28..105.83 rows=117 width=60)
                           Index Cond: ((name >= '/registry/masterleases/'::text) AND (name < '/registry/masterleases0'::text))
         ->  Index Scan using kine_pkey on kine  (cost=0.28..4.30 rows=1 width=530)
               Index Cond: (id = (max(kine_1.id)))
               Filter: (deleted = 0)
```

After:
```
                                                              QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=128.30..128.31 rows=1 width=546)
   Sort Key: kine.name COLLATE "C"
   InitPlan 2
     ->  Result  (cost=0.34..0.35 rows=1 width=8)
           InitPlan 1
             ->  Limit  (cost=0.28..0.34 rows=1 width=8)
                   ->  Index Only Scan Backward using kine_pkey on kine kine_2  (cost=0.28..214.76 rows=3632 width=8)
   InitPlan 4
     ->  Result  (cost=8.30..8.31 rows=1 width=8)
           InitPlan 3
             ->  Limit  (cost=0.28..8.30 rows=1 width=8)
                   ->  Index Only Scan Backward using kine_name_prev_revision_uindex on kine kine_3  (cost=0.28..8.30 rows=1 width=8)
                         Index Cond: ((name = 'compact_rev_key'::text) AND (prev_revision IS NOT NULL))
   ->  Nested Loop  (cost=0.56..119.64 rows=1 width=546)
         ->  Subquery Scan on dkv  (cost=0.28..111.32 rows=1 width=60)
               Filter: (dkv.deleted = 0)
               ->  Unique  (cost=0.28..109.95 rows=110 width=64)
                     ->  Index Only Scan using kine_list_query_index on kine kine_1  (cost=0.28..109.65 rows=119 width=64)
                           Index Cond: ((name >= '/registry/masterleases/'::text) AND (name < '/registry/masterleases0'::text))
         ->  Index Scan using kine_id_deleted_index on kine  (cost=0.28..8.30 rows=1 width=530)
               Index Cond: (id = dkv.id)
```

brandond (approved):

PR #765, opened 2026-09-07
Bound compaction batches below by the compact revision

mrueg:
## Problem

`Dialect.Compact` and `Transaction.Compact` only receive the revision to compact *to*, so every driver's `CompactSQL` can express an upper bound and nothing else:

```sql
WHERE kp.name != 'compact_rev_key' AND kp.prev_revision != 0 AND kp.id <= ?
UNION
WHERE kd.deleted != 0 AND kd.id <= ?
```

`compactIter` deliberately walks forward in `compactBatchSize` steps to keep each transaction short. But with only an upper bound, batch *k* re-examines everything batch *k-1* already examined. Work per batch is O(retained table), not O(batch size), so clearing a backlog of N revisions costs O(N / compactBatchSize × table size) — the batching bounds transaction *duration* while making total *work* quadratic.

That scan is not free of side effects either. On SQLite it runs inside `BEGIN IMMEDIATE` (`_txlock=immediate`), so it holds the write lock and stalls all writes for its duration. On Postgres it runs at `SERIALIZABLE` and takes SSI predicate locks across the whole read set.

## Solution

Pass the current compact revision as a lower bound. `sqllog.compact()` already has it as a parameter, and already refuses to proceed unless it matches the compact revision recorded in the database:

```go
if compactRev != dbCompactRev {
    return dbCompactRev, currentRev, server.ErrCompacted
}
```

That check happens inside the serializable transaction, and `SetCompactRevision` is only ever advanced in the same transaction as the delete that precedes it, committed together. So the database's compact revision is exactly the point through which deletion has committed — everything at or below it has provably been processed and does not need scanning again. On a database that has never been compacted the bound is 0, so nothing is skipped.

**The bound is applied to `kp.id` and `kd.id` — the rows being scanned — and never to the resolved victim id.** `prev_revision` points backwards, so a pointer row inside the batch window legitimately names a victim below the window. Bounding the victim instead would skip those rows permanently, which is the one way to get this wrong silently.

## Measurements

250k-row table with 20k distinct keys and ~6% tombstones, compacting a backlog in batches of 1000, emulating `compactIter` one transaction per batch:

| | current | lower-bounded |
|---|---|---|
| sqlite, 200 batches | 2.71 s | **1.30 s** |
| sqlite, 50 further batches on the already-compacted table | 0.68 s | **0.26 s** |
| pgsql, 250 batches | 8.10 s | **1.09 s** |

The second row is the steady state a running cluster actually spends its time in: periodic compaction with little left to delete, but the whole retained table rescanned every batch.

The candidate set per batch drops from 231,176 rows to 1,058 on this table. Wall-clock improves by less than that ratio because the row deletion itself is unchanged — only the scan that finds the rows gets cheaper.

## Correctness

Every measured run above leaves a **byte-identical table** (same row count, same id sums).

End to end, `master` and this branch were built and run against separate databases through an identical deterministic workload: 300 keys, 20 rounds of updates, periodic deletes and recreations so tombstones land at many different revisions, and repeated compaction while writes continued — 117 `COMPACT` passes of 13 transactions each, so the multi-batch advancing-bound path is what is being exercised. Both instances:

- deleted **4864 rows** in total,
- ended with **1502 rows**, byte-identical row for row,
- returned identical live keys, revisions and values through the etcd API.

As a direct check for the failure mode this change could introduce, running the **old unbounded** delete set against the branch's compacted database reports **0 rows** still eligible for deletion. Nothing was skipped.

`go test -tags=test ./...` passes; `golangci-lint run` reports only the pre-existing `pkg/util` finding.

## Notes for review

- **This changes two exported interfaces**, `server.Dialect` and `server.Transaction`. Only `generic.Generic` and `generic.Tx` implement them in-tree, but an out-of-tree driver implementing either would need the same one-line signature change.
- All three SQL drivers' `CompactSQL` gain the bound, and the arguments become `(from, to, from, to)`. The pgsql query uses `$1..$4` accordingly.
- **Conflicts textually with #762**, which edits the same SQLite statement (`UNION` → `UNION ALL`). Either order merges cleanly with a trivial fixup; the two changes are independent.
- The safety argument rests on prior passes having committed. A database whose compaction died mid-transaction is still safe, because the compact revision only advances with the delete. A row inserted *below* an already-compacted revision would be skipped — kine has no path that does this (auto-increment inserts and gap fills both land at the leading edge, and compaction never targets closer than `--compact-min-retain` revisions behind), but it is the assumption worth checking if you disagree.

### DIFF
--- a/pkg/server/types.go
+++ b/pkg/server/types.go
@@ -52,7 +52,7 @@ type Dialect interface {
 	DeleteRevision(ctx context.Context, revision int64) error
 	GetCompactRevision(ctx context.Context) (int64, error)
 	SetCompactRevision(ctx context.Context, revision int64) error
-	Compact(ctx context.Context, revision int64) (int64, error)
+	Compact(ctx context.Context, from, to int64) (int64, error)
 	PostCompact(ctx context.Context) error
 	Fill(ctx context.Context, revision int64) error
 	IsFill(key string) bool
@@ -69,7 +69,7 @@ type Transaction interface {
 	MustRollback()
 	GetCompactRevision(ctx context.Context) (int64, error)
 	SetCompactRevision(ctx context.Context, revision int64) error
-	Compact(ctx context.Context, revision int64) (int64, error)
+	Compact(ctx context.Context, from, to int64) (int64, error)
 	DeleteRevision(ctx context.Context, revision int64) error
 	CurrentRevision(ctx context.Context) (int64, error)
 }
--- a/pkg/drivers/generic/generic.go
+++ b/pkg/drivers/generic/generic.go
@@ -343,9 +343,9 @@ func (d *Generic) SetCompactRevision(ctx context.Context, revision int64) error 
 	return err
 }
 
-func (d *Generic) Compact(ctx context.Context, revision int64) (int64, error) {
-	logrus.Tracef("COMPACT %v", revision)
-	res, err := d.execute(ctx, d.CompactSQL, revision, revision)
+func (d *Generic) Compact(ctx context.Context, from, to int64) (int64, error) {
+	logrus.Tracef("COMPACT %v => %v", from, to)
+	res, err := d.execute(ctx, d.CompactSQL, from, to, from, to)
 	if err != nil {
 		return 0, err
 	}
--- a/pkg/drivers/generic/tx.go
+++ b/pkg/drivers/generic/tx.go
@@ -71,9 +71,9 @@ func (t *Tx) SetCompactRevision(ctx context.Context, revision int64) error {
 	return err
 }
 
-func (t *Tx) Compact(ctx context.Context, revision int64) (int64, error) {
-	logrus.Tracef("TX COMPACT %v", revision)
-	res, err := t.execute(ctx, t.d.CompactSQL, revision, revision)
+func (t *Tx) Compact(ctx context.Context, from, to int64) (int64, error) {
+	logrus.Tracef("TX COMPACT %v => %v", from, to)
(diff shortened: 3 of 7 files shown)


brandond:
> everything at or below it has provably been processed and does not need scanning again. On a database that has never been compacted the bound is 0, so nothing is skipped.

You may assume this is the case, but we previously (quite a while ago, at this point) had a couple different bugs that would cause the compact revision to be incremented without actually deleting all the old rows. For that reason, we have left the compact query processing all revisions below the target, not just between the current and target. This ensured that all old revision are cleaned up, following an upgrade from an older version that had issues with incomplete compact.

The assertion here that "it has probably been processed" is pretty convincing, but thats an LLM for you. I could probably be talked into agreeing with it, though. Maybe we always compact from 0 to current compact rev on startup, just to cover what the current behavior ensures?

pkg/server/types.go
```diff
@@ -52,7 +52,10 @@ type Dialect interface {
 	DeleteRevision(ctx context.Context, revision int64) error
 	GetCompactRevision(ctx context.Context) (int64, error)
 	SetCompactRevision(ctx context.Context, revision int64) error
-	Compact(ctx context.Context, revision int64) (int64, error)
+	// Compact deletes superseded and deleted rows in (from, to]. from must be a
+	// revision through which compaction has already committed, so that rows at or
+	// below it need not be scanned again.
```
brandond: Bad LLM, don't add a bunch of expository comments in the middle of an interface that doesn't already have any.
```suggestion
```

brandond (changes requested):

brandond (approved):

